PR #24, opened 2022-08-20
SW-1446 Update github workflow to publish npm package 

karthikbtf:
- publish package to npm in staging and production if there's a version change
- left semantics open to support other artifact repositories such as AWS


.github/scripts/compare-artifact-package-versions.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# TODO: add support for AWS Code artifact
+if [ $ARTIFACT_REGISTRY_PROVIDER = 'npm' ]; then
+
+  echo "Checking npm registry for current version"
+  NPM_PACKAGE_VERSION=`npm show @terraware/web-components version`
+  echo "Got npm package version $NPM_PACKAGE_VERSION"
+
+  echo "Checking version in git package.json"
+  PACKAGE_JSON_VERSION=`node -e "process.stdout.write(require(__dirname + '../../dist/package.json').version)"`
```
constanzauanini, line 15: Do we update this version manually? If we make a change in a component for example?

karthikbtf: For now yes. We can evolve this over time.

.github/scripts/compare-artifact-package-versions.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "Package is up to date"
+  else
+    echo "Package needs update"
+    echo "UPDATE_NPM_PACKAGE=true" >> $GITHUB_ENV
+  fi
+
+else
+
+  echo "Nothing to do"
+  exit
```
sgrimm, line 28: You don't really need the `exit` here, but if you do, you might want to make this `exit 0`. Otherwise it will exit with the exit status of the most recent command. Granted, in this case that'll be the `echo` command, but better to be explicit.

.github/workflows/develop.yml
```diff
@@ -13,10 +13,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Set Environment
+        run: ./.github/scripts/set-environment.sh
```
sgrimm, line 17: Don't need the `./` at the beginning of the path.

.github/workflows/develop.yml
```diff
@@ -25,5 +29,19 @@ jobs:
...
+      - name: build
+        run: yarn build-dist
+
+      - name: license report
+        run: yarn license-report
+
+      - name: compare artifact package version
+        run: ./.github/scripts/compare-artifact-package-versions.sh
+        env:
+          ARTIFACT_REGISTRY_PROVIDER: ${{ env.ARTIFACT_REGISTRY_PROVIDER }}
```
sgrimm, line 41: I don't think you need this -- I believe any environment variables defined in the env file that `$GITHUB_ENV` points to will automatically be made available to all `run` jobs.

sgrimm (approved):
A couple minor cosmetic things you might want to tweak, but looks good generally. This is great to get in place!

PR #25, opened 2022-08-22
SW-1468 Update web-components with latest changes

karthikbtf:
- also use new mui
- update distribution to include transpiled .js (from typescript)
- update distribution to include .d.ts generated types


src/components/DatePicker.tsx
```diff
@@ -1,67 +1,44 @@
...
-      size='small'
-      fullWidth={true}
-      className={classes.root}
-      onKeyPress={props.onKeyPress}
-      maxDate={props.maxDate ?? undefined}
-      error={props.error}
-      helperText={props.helperText}
-      disabled={props.disabled}
-    />
+    <LocalizationProvider dateAdapter={AdapterDateFns}>
```
karthikbtf, line 31: @constanzauanini I introduced this change as mui doc show this https://mui.com/x/react-date-pickers/date-picker/ so we don't have all clients introducting localization code.

style-dictionary/build.js
```diff
@@ -5,7 +5,7 @@ function build(source, destination) {
       scss: {
         transformGroup: 'scss',
         transforms: ['name/css_composite', 'value/gradient', 'value/shadow'],
-        buildPath: '../style-dictionary-dist/',
+        buildPath: '../src/style-dictionary-dist/',
```
karthikbtf, line 8: @constanzauanini i moved this into src/ directory because we need this packaged into the distribution

karthikbtf (commented):
@constanzauanini these are 95% copied over files with a few changes.

src/components/CancelButton.tsx
```diff
@@ -1,20 +1,19 @@
-import { Chip } from '@material-ui/core';
-import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
+import { Chip } from '@mui/material';
```
constanzauanini, line 1: I think we should delete this component and just replace it with the new component using a cancel label

src/components/DatePicker.tsx
constanzauanini: Ok! In terraware-web we should delete all the LocalizationProviders that I added

src/components/DialogCloseButton.tsx
```diff
@@ -1,18 +1,17 @@
-import { IconButton } from '@material-ui/core';
```
constanzauanini, line 1: Same here, if we are using the new DialogBox component everywhere, we shouldn't need this CloseButton anymore

src/components/TextField.tsx
```diff
@@ -1,13 +1,13 @@
-import { createStyles, makeStyles, TextField as MUITextField } from '@material-ui/core';
+import { TextField as MUITextField } from '@mui/material';
```
constanzauanini, line 1: Are we removing this component later?

src/components/Textfield/styles.scss
```diff
@@ -1,4 +1,6 @@
-@import '../../../style-dictionary-dist/textfield_variables.scss';
+@use 'sass:math';
```
constanzauanini, line 1: I see we are using math in the button component for example and not having this @use  what is it for?

karthikbtf: This was just copied over.
I believe sass:math is needed fi we are using math otherwise you get warnings.
Cleaning up css is not the scope of this PR.

src/components/DialogCloseButton.tsx
karthikbtf: This and CancelButton and TextField are still being used in terraware-web.
Perhaps something I can use your help cleaning up.
For now I will keep these in the PR, can't do it all in one step.

src/components/CancelButton.tsx
karthikbtf: I don't want this PR to generate a ton of new work. Scope of this PR is to carry over stuff from terraware-web and use them there.
We can clean this up later.
It would be great if we can create a new Jira to capture all the stuff that needs to get cleaned up?

karthikbtf:
@constanzauanini the scope of this PR for now is to keep parity with terraware-web and just carry over the stuff so it can be pulled in as a dependency.
Yes, we need to clean up a lot of component usage, lets not make that scope of this PR - we should create a new Jira for that.

src/components/DialogCloseButton.tsx
constanzauanini: Sure!

karthikbtf: Thanks 🙏🏻 I’ll go ahead and remove them from this PR. You are correct we don’t need to copy them here.

src/components/CancelButton.tsx
karthikbtf: Note: ask Chudo for zeroheight and create a JIRA for cancel label.

karthikbtf:
@constanzauanini I removed CancelButton, DialogCloseButton, TextArea and TextField.
I missed copying over changes from Select, which I just did and confirmed storybook works.

constanzauanini:
> @constanzauanini I removed CancelButton, DialogCloseButton, TextArea and TextField. I missed copying over changes from Select, which I just did and confirmed storybook works.

LGTM!

karthikbtf:
@constanzauanini I updated README on how to add a new .svg file.
We have to convert them to a react component in order for distribution for now.
Will get this in for now before I add any more to this.

karthikbtf:
btw, i tested an integration of this in terraware-web and it is working nicely. Will send a corresponding PR for that tomorrow.

PR #26, opened 2022-08-23
Fix web-components package.json to publish all files in the distribution.

karthikbtf:
The `files` property limited publishing to `index.js` and `package.json` alone and skipped rest of the files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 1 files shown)


sgrimm (changes requested):
I'll grant I know nothing about npm packaging, but I just Googled this as a sanity check and based on https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files it seems like leaving `files` out will package up our `.github` directory which I don't think we'd want.

Is the problem maybe that the `dist` directory didn't get populated before the packaging step?

karthikbtf:
> I'll grant I know nothing about npm packaging, but I just Googled this as a sanity check and based on https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files it seems like leaving `files` out will package up our `.github` directory which I don't think we'd want.
> 
> Is the problem maybe that the `dist` directory didn't get populated before the packaging step?

The dist directory is/was populated. The script currently changes directory into `dist`, objective is to publish all the files in the dist directory. The dist directory is not part of the github repo.
Let me dig around some more, there might be another way to do this.

sgrimm (approved):
Oh, okay, I didn't realize the script changed directories first. If this is running from within `dist`, my concern goes away.

karthikbtf:
> Oh, okay, I didn't realize the script changed directories first. If this is running from within `dist`, my concern goes away.

You brought up a good point. I can publish without changing directories. I will explore doing that in a follow up review, requires setting up `.npmrc` to ignore all the github content (as described in the link you provided). That would be less hacky / more cleaner.

I will do that once I have CodeArtifact set up so I'm not proliferating versions with my testing.

I have to bump the version here one time due to an accidental publish.

PR #28, opened 2022-08-24
Improve exported dependencies

constanzauanini:


karthikbtf (commented):
export from `src/index.ts`
Can we rename Order to SortOrder.
Also the export list should be sorted.. i wanted to do that but missed it.

src/index.ts
```diff
@@ -56,3 +57,4 @@ export {
   getComparator,
   stableSort,
 };
+export type { TableRowType, EnhancedTableDetailsRow, RendererProps, TableColumnType, SortOrder };
```
karthikbtf, line 60: Can we move this to line 32 just to group all the type exports.
If you can, can we sort all the exports?

karthikbtf (commented):
I think you need to make more changes related to SortOrder.
I recommend running `yarn lint`, `yarn build-dist` and `yarn storybook` to catch errors.

src/index.ts
constanzauanini: I can't group them, the linter says I have to add the word "type" to export a type `Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'`

But I will sort them in the two groups

karthikbtf: i meant to say have all the export types appear together, just move it below the previous export type lines.

karthikbtf:
```
export type { IconName } from './components/Icon/icons/';
export type { Size } from './components/Size';
export type { TableRowType, EnhancedTableDetailsRow, RendererProps, TableColumnType, SortOrder };
```

karthikbtf (approved):
looks good, we can bump the package.json version and check this in

PR #29, opened 2022-08-24
Add ErrorBox to web-components

karthikbtf:
- Added ErrorBox as a component.
- Refactored directory of svg components, they are also now exported from index.js since we need `Logo`.

ErrorBox - Default ⋅ Storybook 2022-08-24 14-13-56


src/stories/ErrorBox.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default {
+  title: 'ErrorBox',
+  component: ErrorBox,
+};
+
+const Template: Story<ErrorBoxProps> = (args) => {
+  return (
+    <ErrorBox
+      {...args}
+      text={'This is an error message'}
```
constanzauanini, line 14: Yo don't need to pass this props here because you are already passing them by doing {...args}. Also, if you change values on the fly on storybook they are not going to change if you have them harcoded here

karthikbtf: Thanks, makes sense... will fix.

constanzauanini (approved):

PR #33, opened 2022-08-26
Fix empty table pagination

alexTerraformation:
Empty table pagination used to say "1 to 0 of 0", now it says "0 of 0"
Screen Shot 2022-08-26 at 8 28 54 AM

SW-1323


constanzauanini (approved):
Upgrade the package version before merging

PR #36, opened 2022-08-29
Remove snackbar from web-components Message component

karthikbtf:


src/components/Message/styles.scss
```diff
@@ -21,6 +21,11 @@
   display: flex;
 
   width: 480px;
+
+  &.tw-message--mobile {
+    width: 350px;
```
constanzauanini, line 26: I think for mobile, size is full width, why 350?

constanzauanini (approved):

PR #37, opened 2022-08-30
SW-1541 Support complex objects in autocomplete

karthikbtf:
Today autocomplete only supports strings.
Extend this to support complex objects with custom rendering.
Renamed old Autocomplete to SimpleAutocomplete (to support basic string based autocomplete).

SimpleAutocomplete - Simple Autocomplete Story ⋅ Storybook 2022-08-30 12-08-42

Autocomplete - Autocomplete Story ⋅ Storybook 2022-08-30 12-08-24

### DIFF
--- a/src/stories/SimpleAutocomplete.stories.tsx
+++ b/src/stories/SimpleAutocomplete.stories.tsx
@@ -11,7 +11,7 @@ export default {
   component: SimpleAutocomplete,
 };
 
-const SimpleAutocompleteTemplate: Story<SimpleAutocompleteProps> = (args) => {
+const Template: Story<SimpleAutocompleteProps> = (args) => {
   const [selected, setSelected] = React.useState('');
   const handleChange = (id: string, value: string) => {
     action('onChange')(value);
@@ -21,9 +21,9 @@ const SimpleAutocompleteTemplate: Story<SimpleAutocompleteProps> = (args) => {
   return <SimpleAutocomplete {...args} selected={selected} onChange={handleChange} />;
 };
 
-export const SimpleAutocompleteStory = SimpleAutocompleteTemplate.bind({});
+export const Default = Template.bind({});
 
-SimpleAutocompleteStory.args = {
+Default.args = {
   id: '1',
   label: 'Test',
   options: ['Test 1', 'Test 2', 'Hello'],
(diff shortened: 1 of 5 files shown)


karthikbtf:
This is needed for new species lookup, also this is a nice thing to have in general for any autocomplete.

src/stories/SimpleAutocomplete.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const [selected, setSelected] = React.useState('');
+  const handleChange = (id: string, value: string) => {
+    action('onChange')(value);
+    setSelected(value);
+  };
+
+  return <SimpleAutocomplete {...args} selected={selected} onChange={handleChange} />;
+};
+
+export const SimpleAutocompleteStory = SimpleAutocompleteTemplate.bind({});
```
constanzauanini, line 24: Can we change this 'SimpleAutocompleteStory' to something like 'Default'? This is the name that shows under 'SimpleAutocomplete'

karthikbtf: will do, i had the 2 autocomplete widgets in the same story and renamed them. forgot to switch back to Default when i split the stories.

constanzauanini: Thanks!

constanzauanini (approved):
This looks good! But this autocomplete component is old (with the label that is animated and stay at the top of the input). What I did for new species autocomplete was use the new Select component. I think we should do the change in that component and remove this old autocomplete components

karthikbtf:
> 

Yikes, I guess I wasted time on this :( ... oh well, ok I'll take a look at the Select stuff. Maybe we don't need this PR then. Thanks for letting me know, I did observe that label thing but wasn't sure.

constanzauanini:
> > 
> 
> Yikes, I guess I wasted time on this :( ... oh well, ok I'll take a look at the Select stuff. Maybe we don't need this PR then. Thanks for letting me know, I did observe that label thing but wasn't sure.

I should have mentioned this before :(

karthikbtf:
> > > 
> > 
> > 
> > Yikes, I guess I wasted time on this :( ... oh well, ok I'll take a look at the Select stuff. Maybe we don't need this PR then. Thanks for letting me know, I did observe that label thing but wasn't sure.
> 
> I should have mentioned this before :(

Thanks for catching this, we can remove Autocomplete after we get rid of the v1 accession create/edit.

PR #38, opened 2022-08-30
SW-1541 Add complex objects support to Select component

karthikbtf:
Introduced SelectT (copy of Select with object type made into a generic)

Select - Complex Select ⋅ Storybook 2022-08-30 15-02-27

Select - Default ⋅ Storybook 2022-08-30 15-02-10


src/components/Select/SelectT.tsx
```diff
@@ -0,0 +1,258 @@
+import classNames from 'classnames';
```
karthikbtf, line 1: @constanzauanini this is basically the original `Select` component but string value type has been made a generic `T`.

constanzauanini (approved):
Thanks for adding this support here!

PR #39, opened 2022-08-30
web-components GitHub action to automate design updates PR creation

karthikbtf:
Github action workflow automating design workflow updates + PR creation

### DIFF
--- a/.github/scripts/commit-message.js
+++ b/.github/scripts/commit-message.js
@@ -1,0 +1,9 @@
+const fs = require('fs');
+
+const data = `${process.env.JIRA} ${process.env.DESCRIPTION}\n\n${process.env.DESCRIPTION}`;
+
+fs.writeFile('/tmp/commit-message', data, (err) => {
+  if (err) {
+    process.stderr.write(err);
+  }
+});
--- a/.github/workflows/design-workflow.yml
+++ b/.github/workflows/design-workflow.yml
@@ -4,7 +4,11 @@ on:
   workflow_dispatch:
     inputs:
       description:
-        description: 'What is the change in 5 words or less?'
+        description: 'What are we updating (eg. update button destructive color)?'
+        default: ''
+        required: true
+      jira:
+        description: 'What is the related JIRA (eg. SW-1234)?'
         default: ''
         required: true
       email:
@@ -26,7 +30,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Create branch
-        run: git checkout -b `echo design-workflow-${{ github.actor }}-${{ github.event.inputs.description }} | sed "s/ /-/g"`
+        run: git checkout -b `echo design-workflow-${{ github.actor }}-${{ github.event.inputs.jira }}-\`date +"%y-%m-%d"\`  | sed "s/ /-/g"`
 
       - name: Install Dependencies
         run: yarn install --network-timeout 100000
@@ -51,6 +55,6 @@ jobs:
         run: git push -u origin `git branch --show-current`
 
       - name: Create PR
-        run: gh pr create --fill
+        run: gh pr create --title "${{ github.event.inputs.jira }} - ${{ github.event.inputs.description }}" --body "${{ github.event.inputs.description }}"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/design-workflow.yml
+++ b/.github/workflows/design-workflow.yml
@@ -8,7 +8,7 @@ on:
         default: ''
         required: true
       jira:
-        description: 'What is the related JIRA (eg. SW-1234)?'
+        description: 'What is the related Jira (eg. SW-1234)?'
         default: ''
         required: true
       email:
--- a/.github/workflows/design-workflow.yml
+++ b/.github/workflows/design-workflow.yml
@@ -30,7 +30,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Create branch
-        run: git checkout -b `echo design-workflow-${{ github.actor }}-${{ github.event.inputs.jira }}-\`date +"%y-%m-%d"\`  | sed "s/ /-/g"`
+        run: git checkout -b $(echo design-workflow-${{ github.actor }}-${{ github.event.inputs.jira }}-$(date +%y-%m-%d-%s) | tr ' ' -)
 
       - name: Install Dependencies
         run: yarn install --network-timeout 100000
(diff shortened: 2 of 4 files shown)


.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Create branch
+        run: git checkout -b `echo design-workflow-${{ github.actor }}-${{ github.event.inputs.description }} | sed "s/ /-/g"`
```
karthikbtf, line 29: maybe this branch naming is not a good idea.. using user input

constanzauanini: Maybe we can make user input a branch name?

constanzauanini (approved):

.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      description:
+        description: 'What are we updating (eg. update button destructive color)?'
+        default: ''
+        required: true
+      jira:
+        description: 'What is the related JIRA (eg. SW-1234)?'
```
sgrimm, line 11: Total nitpick: it's "Jira" nowadays (they changed from the all-caps name a few years back).

.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Create branch
+        run: git checkout -b `echo design-workflow-${{ github.actor }}-${{ github.event.inputs.jira }}-\`date +"%y-%m-%d"\`  | sed "s/ /-/g"`
```
sgrimm, line 33: A couple shell things: for command substitution, `$()` nests more cleanly than backticks (no need to escape the nested commands). And `tr` does global character replacement slightly more succinctly than `sed`:
```
run: git checkout -b $(echo design-workflow-${{ github.actor }}-${{ github.event.inputs.jira }}-$(date +%y-%m-%d) | tr ' ' -)
```

.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Build dictionary 
+        run: yarn build-dictionary
+
+      - name: Bump package patch version
+        run: yarn bump patch
+
+      - name: Config user
+        run: |
+          git config --global user.email ${{ github.event.inputs.email }}
```
sgrimm, line 46: I'll be curious to see what this does if a user has more than one email address on their GitHub account.

.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,60 @@
...
+          git config --global user.name ${{ github.actor }}
+
+      - name: Commit changes
+        run: |
+          git add -A
+          git commit -m "${{ github.event.inputs.description }}"
```
sgrimm, line 52: Probably worth testing what this does if the user puts double quotes or dollar signs in the description.

sgrimm (approved):

.github/workflows/design-workflow.yml
karthikbtf: I have 2 emails with my github account (personal email and terraformation email), let me try both emails.

.github/workflows/design-workflow.yml
karthikbtf: Good catch, it does break. 
```
/home/runner/work/_temp/ca28[9](https://github.com/karthikbtf/test/runs/8118436180?check_suite_focus=true#step:9:10)fac-e0b4-478b-b3e9-0ea401b933f2.sh: line 2: unexpected EOF while looking for matching `"'
```
I need to sanitize user input.. hmm,, there's a long list of special shell characters like ! % etc.. let me look online for a comprehensive sanitization code.

sgrimm: If you take that approach, you could write the PR title/description as the first line of the commit message, and then tell `gh pr create` to pull the PR info from the commit message. Otherwise you'll have the same problem with special characters in the `gh` command line, I think.

karthikbtf: > Alternate idea: write the description to a file and use `git commit -F` to read the commit message from there. Should be less funny business with special characters that way.

Let me try that, I have to use some command other than `echo` to write to file.

karthikbtf: Actually echo won't be an issue here.

karthikbtf: File suggestion worked great, using a simple node script to write to file since echo tries to interpret $. Will update this PR.

Screen Shot 2022-08-31 at 10 44 54 AM

.github/workflows/design-workflow.yml
karthikbtf: Both emails were accepted, not sure how this manifests in the PR creation - I see my username in the PRs.

.github/scripts/commit-message.js
```diff
@@ -0,0 +1,9 @@
+const fs = require('fs');
+
+const data = `${process.env.JIRA} ${process.env.DESCRIPTION}\n\n${process.env.DESCRIPTION}`;
```
sgrimm, line 3: What's the benefit of repeating the description?

karthikbtf:
I will get this in for now in order to test - been testing in my private repo so far.
Happy to make follow up changes as needed.

.github/workflows/design-workflow.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Create commit message file
+        run: node .github/scripts/commit-message.js
+        env:
+          JIRA: ${{ github.event.inputs.jira }}
+          DESCRIPTION: ${{ github.event.inputs.description }}
+
+      - name: Commit changes
+        run: |
+          git add -A
+          git commit -F /tmp/commit-message
```
sgrimm, line 58: Oh, good idea to do this with environment variables. In fact, I bet this eliminates the need to write out the file at all:
```
- name: Commit changes
  run: git commit -a -m "$TITLE"
  env:
    TITLE: "${{ github.event.inputs.jira }} ${{ github.event.inputs.description }}"
```

.github/scripts/commit-message.js
karthikbtf: It shows up in the PR description, I agree it is redundant.... I will remove this.

.github/workflows/design-workflow.yml
karthikbtf: Hmm.. good idea, let me give that a shot.

sgrimm: Oops, I guess you need the separate `git add` since there might be new files, so
```
- name: Commit changes
  run: |
    git add -A
    git commit -m "$TITLE"
  env:
    TITLE: "${{ github.event.inputs.jira }} ${{ github.event.inputs.description }}"
```

karthikbtf: Yes, this works great. Follow up PR coming up. Thanks again for the feedback and suggestions.

PR #43, opened 2022-08-31
write info and error to stdout/stderr in download json file

karthikbtf:
@constanzauanini the parallel downloads are failing in github action (i think this needs to be serialized for github action alone). For now just cleaning up errors/info so I can debug further.

### DIFF
--- a/style-dictionary/download-json.js
+++ b/style-dictionary/download-json.js
@@ -35,9 +35,8 @@ const promise = new Promise((resolve, reject) => {
 // Wait for downloads to complete
 return Promise.all(promises)
   .then(() => {
-    console.log('Done downloading all!');
+    process.stdout.write('Done downloading all!');
   })
   .catch((err) => {
-    console.error(err);
-    process.exit(1);
+    process.stderr.write(err);
   });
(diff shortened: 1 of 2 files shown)


constanzauanini (approved):

constanzauanini:
> @constanzauanini the parallel downloads are failing in github action (i think this needs to be serialized for github action alone). For now just cleaning up errors/info so I can debug further.

Do you wanna try the node -parallel again?

karthikbtf:
> > @constanzauanini the parallel downloads are failing in github action (i think this needs to be serialized for github action alone). For now just cleaning up errors/info so I can debug further.
> 
> Do you wanna try the node -parallel again?

Oh good idea, the curl --parallel? I can try that.. wondering if that might fail too since i think it is too many downloads happening too soon. Will try that to confirm.

constanzauanini:
> > > @constanzauanini the parallel downloads are failing in github action (i think this needs to be serialized for github action alone). For now just cleaning up errors/info so I can debug further.
> > 
> > 
> > Do you wanna try the node -parallel again?
> 
> Oh good idea, the curl --parallel? I can try that.. wondering if that might fail too since i think it is too many downloads happening too soon. Will try that to confirm.

Yes, sorry, the curl --parallel. But as you say, maybe it also fails

PR #47, opened 2022-09-01
Upgrade label in datepicker

karthikbtf:
DatePicker - Default ⋅ Storybook 2022-09-01 12-51-30

DatePicker - Default ⋅ Storybook 2022-09-01 12-51-16


constanzauanini (approved):
One thing I think we have to review is the focus/active color of the date picker, I think it's a bit different from the mui active color, but not sure. Maybe @chudo76 can help with that

karthikbtf:
> One thing I think we have to review is the focus/active color of the date picker, I think it's a bit different from the mui active color, but not sure. Maybe @chudo76 can help with that

Yes, we can use design help for that.

PR #49, opened 2022-09-05
Add edit icon

constanzauanini:
Screen Shot 2022-09-05 at 16 28 42


src/components/svg/IconEdit.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
```
karthikbtf, line 1: Actually this file needs to be auto-generated. I added some instructions in the README.
Copy the 'icon-edit.svg' to './assets' and run `yarn generate-assets`.

constanzauanini: Oh, forgot about that!

PR #50, opened 2022-09-06
SW-1558 Autocomplete with new label

karthikbtf:
Autocomplete - Default ⋅ Storybook 2022-09-06 09-06-33

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.14",
+  "version": "1.0.15",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -11,6 +11,7 @@ export interface Props {
   freeSolo: boolean;
   disabled?: boolean;
   className?: string;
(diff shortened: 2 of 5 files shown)


constanzauanini (approved):

karthikbtf:
@constanzauanini this will affect our existing autocompletes as well ( i think we have only one - for species name in create accession ).. it will look like this..
Terraware App 2022-09-06 09-31-45

should i just make this an 'Autocomplete2' or should i use the new Textfield for `Number of Trees` so they are aligned?

karthikbtf:
> @constanzauanini this will affect our existing autocompletes as well ( i think we have only one - for species name in create accession ).. it will look like this.. Terraware App 2022-09-06 09-31-45
> 
> should i just make this an 'Autocomplete2' or should i use the new Textfield for `Number of Trees` so they are aligned?

I added an isV1 option for now, this seems to handle all our use-cases.

constanzauanini:
> > @constanzauanini this will affect our existing autocompletes as well ( i think we have only one - for species name in create accession ).. it will look like this.. Terraware App 2022-09-06 09-31-45
> > should i just make this an 'Autocomplete2' or should i use the new Textfield for `Number of Trees` so they are aligned?
> 
> I added an isV1 option for now, this seems to handle all our use-cases.

Yes, I think this was the best option. Thank you!

PR #59, opened 2022-09-13
SW-1617 Set bgcolor for toast messages

karthikbtf:


constanzauanini (approved):

karthikbtf:
@constanzauanini made one more change to support showing errors in date picker. I tried reusing Textfield since it has built-in error message etc but that didn't work very well. Added error message from scratch.

DatePicker - Default ⋅ Storybook 2022-09-13 10-40-37

constanzauanini (approved):

PR #62, opened 2022-09-21
Scroll bug on Dialog box

constanzauanini:


karthikbtf (approved):
Looks good, thanks for catching this.
Going forward, lets add descriptions to PRs?  In this case the title was self-explanatory but maybe just as a habit, i need to do that as well.

constanzauanini:
> Looks good, thanks for catching this. Going forward, lets add descriptions to PRs? In this case the title was self-explanatory but maybe just as a habit, i need to do that as well.

Sure! I will do it in next prs

PR #63, opened 2022-09-24
SW-1738 add theme in web-components

karthikbtf:
- generate javascript file of theme styles
- merge theme styles into created theme
- export theme in web-components index
- use storybook to demonstrate that this works

Note: the token names are the current ones - this is ok, mostly demonstrates that the process works. We will clean this up once we have newer names.

terraware-web can now import theme from web-components and refer to logical names instead of using hard coded colors.

Theme - Default ⋅ Storybook 2022-09-23 17-37-55


karthikbtf:
@constanzauanini when you get a chance let me know your thoughts on this.
The process/scripts will mostly be the same once we have the new token names. 
The only place with most changes will be terraware-web - we could wait until we get the new token names before making changes there - or we could start with these old names and switch to new names once they are ready.

src/style-dictionary-dist/TerrawareTheme.ts
```diff
@@ -0,0 +1,831 @@
...
+    "ClrDvdrFill": "#A9B7B8",
+    "ClrIconFillCritical": "#FE0003",
+    "ClrIconFillPassive": "#708284",
+    "ClrIconFillProductive": "#007DF2",
+    "ClrIconFillSuccess": "#308F5F",
+    "ClrIconFillText": "#3A4445",
+    "ClrIconFillTextInv": "#FFFFFF",
+    "ClrIconFillWarning": "#BD6931",
+    "ClrBaseBlack": "#000000",
+    "ClrBaseGray025": "#F9FAFA",
```
constanzauanini, line 540: Are we going to have all the color in pallete and in terraware also?

karthikbtf: for now yes, i will clean it up once we get the new tokens - since they won't be starting with Clr.

constanzauanini:
> @constanzauanini when you get a chance let me know your thoughts on this. The process/scripts will mostly be the same once we have the new token names. The only place with most changes will be terraware-web - we could wait until we get the new token names before making changes there - or we could start with these old names and switch to new names once they are ready.

This looks really good. We can now get all token from the theme. Excellent!! I think it is going to be easier if we wait for the new tokens names, so we don't have to change colors two times in terraware-web

constanzauanini (approved):

karthikbtf:
> > @constanzauanini when you get a chance let me know your thoughts on this. The process/scripts will mostly be the same once we have the new token names. The only place with most changes will be terraware-web - we could wait until we get the new token names before making changes there - or we could start with these old names and switch to new names once they are ready.
> 
> This looks really good. We can now get all token from the theme. Excellent!! I think it is going to be easier if we wait for the new tokens names, so we don't have to change colors two times in terraware-web

My thought is we can import this theme and maybe replace just one color in terraware-web - just as a proof of concept to show that this process of importing the theme works. When we get the new tokens we can update all the colors - there will be a little less work (since the theme is already imported, not that's it's a lot of work). Does that sound ok?

constanzauanini:
> > > @constanzauanini when you get a chance let me know your thoughts on this. The process/scripts will mostly be the same once we have the new token names. The only place with most changes will be terraware-web - we could wait until we get the new token names before making changes there - or we could start with these old names and switch to new names once they are ready.
> > 
> > 
> > This looks really good. We can now get all token from the theme. Excellent!! I think it is going to be easier if we wait for the new tokens names, so we don't have to change colors two times in terraware-web
> 
> My thought is we can import this theme and maybe replace just one color in terraware-web - just as a proof of concept to show that this process of importing the theme works. When we get the new tokens we can update all the colors - there will be a little less work (since the theme is already imported, not that's it's a lot of work). Does that sound ok?

Yes, good idea! 👍

karthikbtf:
I messed up, I have to add the palette type definition as well - so we can refer to the color/token names. will do that next. Sorry for this intermediate version that does nothing.

PR #65, opened 2022-09-27
Export types for terraware theme in web-components

karthikbtf:
- the types are needed for typescript to succeed parsing `theme.palette.<terraware-token-name>`
- also remove non-palette colors

### DIFF
--- a/style-dictionary/build-theme.js
+++ b/style-dictionary/build-theme.js
@@ -15,12 +15,41 @@ function build(source, destination) {
     },
   });
 
+  const PALETTE_TYPE_DEFINITION = `import { PaletteOptions } from '@mui/material/styles/createPalette';
+
+declare module '@mui/material/styles/createPalette' {
+  export interface Palette {
+<TERRAWARE_PALETTE_TYPE_DEFINITION>
+  }
+
+  export interface PaletteOptions {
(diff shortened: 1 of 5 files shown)


style-dictionary/build-theme.js
```diff
@@ -15,12 +15,41 @@ function build(source, destination) {
...
+  }
+}`;
+
+const TERRAWARE_PALETTE = `type ThemeDictionary = {
+  [key: string]: any;
+};
+
+const TerrawareTheme: ThemeDictionary = <TERRAWARE_PALETTE>
+
+export default TerrawareTheme;`;
```
constanzauanini, line 36: Is this _**`;**_ and error?

karthikbtf: Nope,, that's the string that will be written to the output file:

```
export default TerrawareTheme;
```

karthikbtf: The second ; is to terminate the current string value - which internally contains another ; which will end up in the file.

constanzauanini (approved):

style-dictionary/build-theme.js
constanzauanini: Ok, thanks for the explanation

PR #68, opened 2022-09-29
SW-1590 Clean up icon options for buttons

alexTerraformation:


src/components/Button/Button.tsx
```diff
@@ -4,6 +4,16 @@ import { IconName } from '../Icon/icons';
 import { Size } from '../Size';
 import './styles.scss';
 
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles(() => ({
+  svgIconFill: {
+    '& path': {
+      fill: 'currentColor',
```
constanzauanini, line 12: What does this style do? Can we add it in the .scss file?

karthikbtf: This is needed to override the fill in some of our icon assets (some icon-* assets have a path fill embedded in the svg).
There's a ticket out for Chudo to clean them up - until then we need these overrides.

karthikbtf: To clarify, ideally this should be a `fill = color` but because of the embedded style in svg, we need to override with the `&path` qualification.

alexTerraformation: Yes, thanks @karthikbtf.  So in this case it makes the svg path fill the same color as the button text.

constanzauanini: Oh, ok. thanks for the explanation

constanzauanini (approved):

src/components/Button/Button.tsx
karthikbtf: i misread @constanzauanini 's suggestion, moving this class to the styles.scss would be ideal.

karthikbtf (commented):
Looks good. Also need to bump the version in package.json.

karthikbtf (approved):

PR #69, opened 2022-09-30
SW-1831 add white background to datepicker text input

karthikbtf:
- currently uses parent background
- updated storybook to capture this behavior

DatePicker - Default ⋅ Storybook 2022-09-30 15-14-04


karthikbtf:
@constanzauanini in case you missed this

constanzauanini (approved):

PR #70, opened 2022-10-04
SW-1919  allow custom styles in table cell/header

karthikbtf:
- updated table story to demonstrate changes

### DIFF
--- a/src/components/table/TableCellRenderer.tsx
+++ b/src/components/table/TableCellRenderer.tsx
@@ -54,7 +54,7 @@ export function CellDateRenderer({
   const classes = useStyles();
 
   return (
-    <TableCell id={id} align='left' className={classes.date + ' ' + (className || '')}>
+    <TableCell id={id} align='left' className={classes.date + ' ' + className}>
       <Typography component='p' variant='body1'>
         {cellDateFormatter(value)}
       </Typography>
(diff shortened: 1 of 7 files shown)


src/components/table/TableHeader.tsx
```diff
@@ -33,7 +34,7 @@ export default function EnhancedTableHead(props: Props): JSX.Element {
   }, [props.columns]);
 
   return (
-    <thead>
+    <TableHead>
```
karthikbtf, line 37: @constanzauanini do you know why we had `thead` before?
`TableHead` is required for `stickyHeader`

constanzauanini: I think it was for reording the table heads, does it still work?

karthikbtf: stickyheader works here in story example but not on web-components.. looks like we need more changes.
i will remove stickyheader changes.. lets just focus on the overriding styles in table header and cells.

karthikbtf: we need a max-height on table container with overflow auto.. our terraware-web views are a bit tricky for max-height at this point.

karthikbtf: @constanzauanini i removed sticky header fixed for now - scope of this PR is just the overriding styles

src/components/table/TableCellRenderer.tsx
```diff
@@ -74,34 +85,52 @@ export function CellTextRenderer({
...
   value,
+  className,
 }: {
   id: string;
   value?: string | number | any[] | ReactNode;
+  className?: string;
 }): JSX.Element {
   return (
-    <TableCell id={id} align='left'>
+    <TableCell id={id} align='left' className={className || ''}>
```
constanzauanini, line 95: I don't think all this ors are necessary. Classname can be undefined

karthikbtf: I added them because we keep seeing undefined in the html markup when inspecting elements. This is just cleaner.

constanzauanini: Oh, ok!

constanzauanini (approved):

PR #71, opened 2022-10-05
SW-1954 don't default number inputs to '0'

karthikbtf:
- this was being harcoded, removed that
- won't default to any value now


charlie-tf (approved):
Looks good to me!

karthikbtf:
> Looks good to me!

Thanks, Charlie!

PR #72, opened 2022-10-06
SW-1837 Better UTC handing in date picker

karthikbtf:
- using moment for most efficient timezone defaults
- added utilities for checking if date is in the future (to be used in terraware-web)
- tested in PDT when UTC was next day
- also tested in terraware-web (was able to select Oct 6 - when Oct 5 in PDT, Oct 7 was rejected by utility check as expected)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -47,22 +47,27 @@ export default function DatePicker(props: Props): JSX.Element {
...
-          maxDate={props.maxDate}
-          inputFormat='yyyy-MM-dd'
+          minDate={props.minDate ? moment(props.minDate) : undefined}
+          maxDate={props.maxDate ? moment(props.maxDate) : undefined}
+          inputFormat='yyyy-MM-DD'
           value={props.value ? moment(props.value, 'YYYY-MM-DD').toDate() : null}
-          onChange={(newValue: string | null) => {
-            props.onChange(props.id, newValue);
+          onChange={(newValue: any) => {
+            if (newValue?.isValid()) {
```
constanzauanini, line 68: Should we show an error if is not valid?

karthikbtf: No the user may be typing a value and it may be incomplete.

karthikbtf: Let me check something.

karthikbtf: https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/15-is-valid/ checks if the date is valid
If user types '2022-' that is not valid

constanzauanini (approved):

src/components/DatePicker/DatePicker.tsx
```diff
@@ -63,11 +64,10 @@ export default function DatePicker(props: Props): JSX.Element {
...
           maxDate={props.maxDate ? moment(props.maxDate) : undefined}
           inputFormat='yyyy-MM-DD'
-          value={props.value ? moment(props.value, 'YYYY-MM-DD').toDate() : null}
+          value={temporalValue}
           onChange={(newValue: any) => {
-            if (newValue?.isValid()) {
-              props.onChange(props.id, newValue?.toDate());
-            }
+            setTemporalValue(newValue);
+            props.onChange(props.id, newValue?.toDate(), newValue.isValid());
```
karthikbtf, line 70: @constanzauanini i made this change so callers of this component can do error handling (like we currently do).

karthikbtf: i reverted this - let the clients decide if date is valid or not.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -47,22 +48,26 @@ export default function DatePicker(props: Props): JSX.Element {
...
-          value={props.value ? moment(props.value, 'YYYY-MM-DD').toDate() : null}
-          onChange={(newValue: string | null) => {
-            props.onChange(props.id, newValue);
+          minDate={props.minDate ? moment(props.minDate) : undefined}
+          maxDate={props.maxDate ? moment(props.maxDate) : undefined}
+          inputFormat='yyyy-MM-DD'
+          value={temporalValue}
+          onChange={(newValue: any) => {
+            setTemporalValue(newValue);
+            props.onChange(props.id, newValue?.isValid() ? newValue?.toDate() : null);
```
constanzauanini, line 70: So we don't change the value if it's not valid, right? Why did you say that client will decide that?

karthikbtf: we send back null to the client if it's not valid - this is the closest we can get to backwards compatibility because moment passes a date object even if date is not correct. 
But moment.isValid() will catch that.

constanzauanini (approved):

PR #74, opened 2022-10-06
SW-1720 Sort rows before splitting into pages

constanzauanini:


karthikbtf (approved):
That was quick, thanks!! 🎉

PR #75, opened 2022-10-11
SW-1811 Adding margin to dialog box on web-mobile

constanzauanini:
Screen Shot 2022-10-11 at 10 29 50


charlie-tf (approved):
ooo token, nice!

PR #83, opened 2022-10-14
Fix IconTooltip Icon Spacing

charlie-tf:
- `box-sizing: border-box` causes padding to be inside the 20x20 box
- this shrinks down the icon to tiny size
- replace padding with margin, which is outside the box
- TODO: figure out how to add `box-sizing: border-box` in
the correct places in Storybook so that things render the same as in 
the terraware-web app.

In the app, this element inherits `box-sizing: border-box`:
image

It's not seen in Storybook, as the `box-sizing` is not applied there:
image

Changing the padding to margin allows that spacing to be outside the 20x20 box of the icon, thereby not crushing the icon glyph:
image

### DIFF
--- a/src/components/IconTooltip/index.tsx
+++ b/src/components/IconTooltip/index.tsx
@@ -14,8 +14,8 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   icon: {
     fill: theme.palette.ClrBaseGray800,
-    paddingLeft: '.5em',
-    paddingRight: '.5em',
+    marginLeft: '.5em',
+    marginRight: '.5em',
     verticalAlign: 'text-top',
   },
   tooltip: {


joshbuchea:
@charlie-tf thank you for looking into this 🙌

I have another [PR](https://github.com/terraware/web-components/pull/84) open that addresses this and a couple of other issues I discovered yesterday but was too tired to open a PR EOD. Please take a look when you have a chance 🙏

charlie-tf:
> @charlie-tf thank you for looking into this 🙌
> 
> I have another [PR](https://github.com/terraware/web-components/pull/84) open that addresses this and a couple of other issues I discovered yesterday but was too tired to open a PR EOD. Please take a look when you have a chance 🙏

@joshbuchea Thanks! I'm currently trying to figure out how to get the story book to render the same as the app (it's that 'box-sizing' property), but banging my head against it. I think I'll probably set it aside for a bit and close this PR.

PR #85, opened 2022-10-16
SW-1901 Add Ability to Float Table Header

charlie-tf:
- create a useScrollDebounce hook to detect scroll position and call effect after a settling time
- in TableHeader compare scroll position against the position of the table container and compute an offset to move the header by if scrolled below the top of the container
- an additional amount may be added to offset the header from the top of the screen
- smoothly animate moving the header to its new position when it updates
- add a story for the Floating Header case for Table

### DIFF
--- a/src/components/table/index.tsx
+++ b/src/components/table/index.tsx
@@ -1,5 +1,5 @@
 import { Box, Checkbox, Pagination, Table, TableBody, TableCell, TableContainer, TableRow, Theme, TooltipProps, Typography } from '@mui/material';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import EnhancedTableToolbar from './EnhancedTableToolbar';
 import { descendingComparator, getComparator, SortOrder, stableSort } from './sort';
 import TableCellRenderer from './TableCellRenderer';
@@ -61,6 +61,8 @@ export interface Props<T> {
   showPagination?: boolean;
   controlledOnSelect?: boolean;
   reloadData?: () => void;
+  floatHeader?: boolean;
+  headerWindowOffset?: number;
 }
 
 export type TopBarButton = {
@@ -93,6 +95,8 @@ export default function EnhancedTable<T>({
   showPagination = true,
   controlledOnSelect,
   reloadData,
+  floatHeader,
+  headerWindowOffset,
 }: Props<T>): JSX.Element {
   const classes = tableStyles();
   const [order, setOrder] = React.useState<SortOrder>(_order);
@@ -100,6 +104,7 @@ export default function EnhancedTable<T>({
   const [maxItemsPerPage] = useState(100);
   const [itemsToSkip, setItemsToSkip] = useState(0);
   const { isMobile } = useDeviceInfo();
+  const containerRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
     if (setSelectedRows && rows.length >= 0) {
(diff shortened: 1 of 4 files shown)


src/components/table/index.tsx
```diff
@@ -201,7 +206,7 @@ export default function EnhancedTable<T>({
   return (
     <>
       {showTopBar && <EnhancedTableToolbar numSelected={selectedRows ? selectedRows.length : 0} topBarButtons={topBarButtons} />}
-      <TableContainer id={id}>
+      <TableContainer id={id} ref={containerRef}>
         <DndContext sensors={sensors} onDragEnd={handleDragEnd}>
           <Table stickyHeader={true} aria-labelledby='tableTitle' size='medium' aria-label='enhanced table' className={classes.table}>
```
charlie-tf, line 211: I tried forever to get the built-in `stickyHeader` functionality to work. The best I could do involved fixing the height of the table container, which produced a gross double-scrollbar effect.

charlie-tf, line 211: I tried forever to get the built-in `stickyHeader` functionality to work. The best I could do involved fixing the height of the table container, which produced a gross double-scrollbar effect.

src/utils/useScrollDebounce.ts
```diff
@@ -0,0 +1,51 @@
...
+          await new Promise((r) => setTimeout(r, debounceTime));
+          const newPos = getScrollPosition();
+          positionEqual = pos.y === newPos.y && pos.x === newPos.x;
+          pos = newPos;
+        }
+        effect(pos);
+        busy.current = false;
+      }
+    };
+    window.addEventListener('scroll', handleScroll, { passive: true });
```
charlie-tf, line 43: on `{ passive: true }`: https://stackoverflow.com/a/37721906/20167578

src/utils/useScrollDebounce.ts
```diff
@@ -0,0 +1,51 @@
...
+    x: scrollPosition.x || 0,
+    y: scrollPosition.y || 0,
+  };
+};
+
+/**
+ * Detects scroll events in the window and calls the effect with the scroll
+ * position when it has settled for longer than the debounceTime (ms).
+ */
+const useScrollDebounce = (
```
charlie-tf, line 21: This hook was adapted from https://github.com/n8tb1t/use-scroll-position, which allows for a time increment on checking scroll position, but not debouncing.

src/components/table/index.tsx
charlie-tf: However, if anyone has any guidance on how to make that work it could be a lot simpler than what I implemented here, and also hopefully do an actual sticky effect instead of floating in like what mine does.

karthikbtf: This seems to work fine in the vercel preview.

There is another way to make this work if you want to try that and then decide which one might be better...

1) change thead to TableHead (like you have already done) here https://github.com/terraware/web-components/blob/main/src/components/table/TableHeader.tsx#L38 - I don't have enough context on why we are using thead here instead of TableHead, we could check with @constanzauanini to see if she remembers
2) no changes here https://github.com/terraware/web-components/blob/main/src/components/table/index.tsx#L206 (see `stickyHeader={true}` option that supports a sticky header)
3) setting a fixed height on this https://github.com/terraware/web-components/blob/main/src/components/table/index.tsx#L204 (now that's the tricky part.. ) - will enable sticky header to work properly

constanzauanini: I remember thead was related to allow dragging columns to reorder them. If that still works with TableHead, then we can change it. We can test that on accession table, or maybe we should make the table story drag to work, so we can test it there

charlie-tf: With a change to TableHead, dragging and dropping the column headers seems to work, as well as clicking to change the sort.

charlie-tf:
found a simpler solution: https://github.com/terraware/web-components/pull/86

PR #87, opened 2022-10-17
SW-1698 remove unintended fill from svg assets 

karthikbtf:
- we need the fill in some assets, this needs to be manually curated


constanzauanini (approved):

karthikbtf:
tested locally, seems fine

PR #88, opened 2022-10-18
SW-2016 Fix Numerical Sorting

charlie-tf:
- Default comparator first attempts parseFloat on both values
- If successful, then compare resulting numeric values
- If either is NaN, then fall back to default (string) compare
- Add some unit tests for the sort

### DIFF
--- a/src/components/table/sort.test.ts
+++ b/src/components/table/sort.test.ts
@@ -28,15 +28,15 @@ test('compare equal string fields', () => {
 
 test('compare unequal number fields with same whole number digits', () => {
   const cmp = descendingComparator(rowA, rowB, 'number1');
-  expect(cmp).toEqual(1);
+  expect(Math.sign(cmp)).toEqual(1);
 });
 
 test('compare unequal number fields with differing whole number digits', () => {
   const cmp = descendingComparator(rowA, rowB, 'number2');
-  expect(cmp).toEqual(-1);
+  expect(Math.sign(cmp)).toEqual(-1);
 });
 
 test('compare equal number fields', () => {
   const cmp = descendingComparator(rowA, rowB, 'number3');
-  expect(cmp).toEqual(0);
+  expect(Math.sign(cmp)).toEqual(0);
 });
--- a/src/components/table/sort.ts
+++ b/src/components/table/sort.ts
@@ -1,4 +1,4 @@
-export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): 1 | -1 | 0 {
+export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): number {
   // first attempt to parse into a numeric value and compare
   const numCompare = descendingNumComparator(a, b, orderBy);
   if (numCompare !== null) {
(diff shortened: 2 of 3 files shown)


src/components/table/sort.test.ts
```diff
@@ -0,0 +1,42 @@
+import { descendingComparator } from './sort';
```
karthikbtf, line 1: nice, thanks for the test

src/components/table/sort.ts
```diff
@@ -11,6 +19,23 @@ export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): 1 | -1 |
...
+  const bNumValue = parseFloat((b[orderBy] ?? '0.0') as string);
+  if (!isNaN(aNumValue) && !isNaN(bNumValue)) {
+    if (bNumValue < aNumValue) {
+      return -1;
+    }
+    if (bNumValue > aNumValue) {
+      return 1;
+    }
+
+    return 0;
```
karthikbtf, line 33: a shortcut for numerical comparison is (bNumValue - aNumValue)

karthikbtf (approved):

src/components/table/sort.ts
charlie-tf: argh 
image

charlie-tf: I might just leave as is instead of fighting with the typechecker :/

karthikbtf: does not have to be -1
has to be negative for less than, zero for same, positive for greater than

short notation for numerical sort is quite common..

for reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Screen Shot 2022-10-18 at 9 41 14 AM

karthikbtf: > I might just leave as is instead of fighting with the typechecker :/

sure, that's fine.

charlie-tf: I'll have to change the return type for `descendingComparator` to `number`, but i can give that a try and make sure the sort still works in the table 👍

charlie-tf: Oh, it should I guess. `stableSort` already expects `(a: T, b: T) => number` 😅

karthikbtf: > I'll have to change the return type for `descendingComparator` to `number`, but i can give that a try and make sure the sort still works in the table 👍

Yikes, just saw the type in descendingComparator. Did not intend for my suggestion to generate a lot of change.
Don't bother about this change for now - what you have is good and correct.

charlie-tf: ah, read this too late 😂 It's okay I already have the change in and tested it

karthikbtf: Thanks so much!

karthikbtf (approved):

karthikbtf:
@charlie-tf let me get a PR out to handle this conflict.. give me a second

karthikbtf:
> 

if you have a sec https://github.com/terraware/web-components/pull/89/files

charlie-tf:
> > 
> 
> if you have a sec https://github.com/terraware/web-components/pull/89/files

Thanks! Once that merges, I'll pull main

karthikbtf:
> > > 
> > 
> > 
> > if you have a sec https://github.com/terraware/web-components/pull/89/files
> 
> Thanks! Once that merges, I'll pull main

merged, you should be able to rebase now

PR #91, opened 2022-10-20
SW-1817 Create TerrawareTheme2 from New Tokens

charlie-tf:
- create style-dictionary2 with scripts for new tokens
- add new script to yarn build-dictionary command
- run and generate the new scss and theme
- import the new theme and merge with existing and base
- add new tokens to progress spinner styles to check build
- modify the theme story to allow selecting TerrawareTheme2
tokens for the background

### DIFF
--- a/src/stories/Theme.stories.tsx
+++ b/src/stories/Theme.stories.tsx
@@ -17,7 +17,7 @@ const Template: Story<{}> = () => {
   const theme = useTheme();
   const [colorValue, setColorValue] = React.useState(theme.palette.ClrIconFillProductive);
   const [colorStr, setColorStr] = React.useState('');
-  const [bgColorValue, setBgColorValue] = React.useState('');
+  const [bgColorValue, setBgColorValue] = React.useState(theme.palette.TwClrBgBrandTertiary);
   const [bgColorStr, setBgColorStr] = React.useState('');
   const palette = theme.palette as unknown as { [key: string]: string };
 


charlie-tf:
Design Changes Workflow GH Action succeeds except for the PR creation step, which may be because I ran it on this branch instead of main
image

src/components/ProgressCircle/styles.scss
```diff
@@ -8,11 +9,11 @@
 .circle-fill {
   position: absolute;
   left: 0;
-  color: $tw-clr-base-semantic-productive;
+  color: $tw-clr-icn-brand;
```
karthikbtf, line 12: the mappings for these tokens are here https://docs.google.com/spreadsheets/d/1ECFEO33DP0GgesPagynY5MIIx_4lFn3rJh_PWjcfqLI/edit#gid=1168800656
(see Jira https://terraformation.atlassian.net/browse/SW-2005)

src/style-dictionary-dist2/TerrawareTheme2.ts
```diff
@@ -0,0 +1,533 @@
+import { PaletteOptions } from '@mui/material/styles/createPalette';
+
+declare module '@mui/material/styles/createPalette' {
+  export interface Palette {
+    TwClrBaseBlue050?: React.CSSProperties['color'];
```
karthikbtf, line 5: should we maybe make these hyphenated?
tw-clr-base-blue-100?

style-dictionary2/transforms/name_css_colors.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
karthikbtf, line 1: maybe we could create a library of transforms that get pulled into the old and new scripts?

karthikbtf (commented):
@charlie-tf it did not occur to me that non-color tokens would still be imported the old way.. looks like we can't get rid of the old flows yet.
wondering if we need to revisit the parallel approach.
maybe just use a `colors` suffix instead of `2`?

karthikbtf:
> > it did not occur to me that non-color tokens would still be imported the old way.. looks like we can't get rid of the old flows yet. wondering if we need to revisit the parallel approach. maybe just use a colors suffix instead of 2?
> 
> @karthikbtf, will the non-color tokens eventually be replaced? If so, I think it makes sense to keep things in version 2 directories/names like i have here, bring in the tokens as they become available, and delete them from the version 1 directory once their usage has been fully deprecated. What do you think?

I think so yes, but not anytime soon. Hasn't been planned or scoped yet.
If we are ok with having `dir` and `dir2` until that time, that's fine too.

style-dictionary2/transforms/name_css_colors.js
charlie-tf: I can do that, good call

src/style-dictionary-dist2/TerrawareTheme2.ts
charlie-tf: Hmm, that does make sense for scss variables, but camelCase is more standard for TS names isn't it?

charlie-tf:
> > > it did not occur to me that non-color tokens would still be imported the old way.. looks like we can't get rid of the old flows yet. wondering if we need to revisit the parallel approach. maybe just use a colors suffix instead of 2?
> > 
> > 
> > @karthikbtf, will the non-color tokens eventually be replaced? If so, I think it makes sense to keep things in version 2 directories/names like i have here, bring in the tokens as they become available, and delete them from the version 1 directory once their usage has been fully deprecated. What do you think?
> 
> I think so yes, but not anytime soon. Hasn't been planned or scoped yet. If we are ok with having `dir` and `dir2` until that time, that's fine too.

Okay, I think I'll keep it as 2, to indicate succession to a newer version of these.

src/style-dictionary-dist2/TerrawareTheme2.ts
karthikbtf: was thinking about how we might end up using the tokens in terraware-web:
`theme.palette.TwClrBaseBlue050` vs `theme.palette.tw-clr-base-blue-050`
I guess we could start with the way it is here and evolve as we get feedback from all devs

charlie-tf: actually, what you have written is more readable I think. I'll try getting it into that format.

charlie-tf: i.e. for this particular case, of referencing styles, kebab-case is a bit less friction, probably

karthikbtf: i do agree with you on readability and ease-of-use

style-dictionary2/build-theme.js
```diff
@@ -3,7 +3,7 @@ function build(source, destination) {
     source,
     platforms: {
       js: {
-        transformGroup: 'js',
+        transforms: ['attribute/cti', 'name/cti/kebab', 'size/rem', 'color/hex'],
```
charlie-tf, line 6: I used the same transforms from the `js` group, but replaced the `name/cti/pascal` with `name/cti/kebab` to get us that other naming format (see https://amzn.github.io/style-dictionary/#/transform_groups?id=js)

charlie-tf: dang, build does not like kebab
image

src/style-dictionary-dist2/TerrawareTheme2.ts
```diff
@@ -0,0 +1,533 @@
+import { PaletteOptions } from '@mui/material/styles/createPalette';
+
+declare module '@mui/material/styles/createPalette' {
+  export interface Palette {
+    tw-clr-base-blue-050?: React.CSSProperties['color'];
```
karthikbtf, line 5: nice!!

charlie-tf: i like it, too, but it caused the build to fail 😭

karthikbtf: ok so maybe your original idea of PascalCase is safer..
something like `'tw-clr-base-blue-050'?: React.CSSProperties['color'];` might work.. but we may need to  reference it as such `theme.palette['tw-clr-base-blue-050']`.. `theme.palette.tw-clr-base-blue-050` may not work.
wondering if we can control this with a babel and/or tslint config.

karthikbtf: fwiw, `-` is interpreted as the `subtraction operator` in `theme.palette.tw-clr-something` .. hence the need to `theme.palette['tw-clr-something']`
maybe best to stick with PascalCase for now - unless we think it's ok to require such usage `theme.palette['abc-def']`?

charlie-tf: I think sticking with PascalCase seems like the way to go for now 👍

charlie-tf:
> Missed this suggestion earlier: we could extend a theme story to support this new theme. https://github.com/terraware/web-components/blob/main/src/stories/Theme.stories.tsx#L12 uses colors from the theme for a foreground and background color. we could use TerrawareTheme2 for the background colors maybe?

Great suggestion! It showed me that `deepmerge` with three arguments was not working correctly.

karthikbtf:
> > Missed this suggestion earlier: we could extend a theme story to support this new theme. https://github.com/terraware/web-components/blob/main/src/stories/Theme.stories.tsx#L12 uses colors from the theme for a foreground and background color. we could use TerrawareTheme2 for the background colors maybe?
> 
> Great suggestion! It showed me that `deepmerge` with three arguments was not working correctly. I wasn't previously sure how to test it.

looks like the signature is
```
export default function deepmerge<T>(target: T, source: unknown, options?: DeepmergeOptions): T;
```
we should be able to
```
deepMerge(deepMerge(defaultPalette, terrawareTheme), terrawareTheme2);
```

also in the theme story, we can add a way to test the typing by initializing the bgcolor value with a theme2 variable https://github.com/terraware/web-components/blob/main/src/stories/Theme.stories.tsx#L18 here

karthikbtf (approved):
Thank you so much for doing this @charlie-tf 
Looks great 🎉 
I had one last comment on an initialization in the theme story - feel free to merge.

PR #92, opened 2022-10-21
Add TableContainer Back

charlie-tf:
- turned out that removing TableContainer was not necessary
- put the TableContainer back in
- add overflowX visible to make the header sticky


constanzauanini (approved):
Oh, this change looks clean! Great that is working

PR #106, opened 2022-10-25
Styling updates for textfields and buttons

alexTerraformation:
Add import and export icons.  Make icon only secondary buttons round.  Make focus state the same as active state for textfields.

Screen Shot 2022-10-25 at 9 25 19 AM
Screen Shot 2022-10-25 at 9 26 40 AM


src/components/Select/styles.scss
```diff
@@ -159,8 +171,13 @@
       }
 
       &:active {
-        border: $tw-sz-frm-fld-text-input-stroke solid $tw-clr-brdr-warning;
-        outline: $tw-sz-frm-fld-text-input-stroke solid $tw-clr-brdr-warning;
+        border: math.div($tw-sz-frm-fld-text-input-stroke-active, 2) solid $tw-clr-brdr-warning;
+        outline: math.div($tw-sz-frm-fld-text-input-stroke-active, 2) solid $tw-clr-brdr-warning;
+      }
+
+      &:focus-within {
```
constanzauanini, line 178: Are you using focus-within instead of focus for any particular reason?

constanzauanini:
Looks good! Remember to bump up the version

src/components/Select/styles.scss
alexTerraformation: The focus is applied to the input within the text field.  Need to apply the styles to the textfield.

constanzauanini: Excellent!

constanzauanini (approved):

PR #107, opened 2022-10-25
SW-1860 Truncate and Show More List of Strings

charlie-tf:
- create a component that shows a list of string
- given a max length, truncate the list
- display a link button indicating the remaining count
- clicking the link button shows a tooltip with the full list

image
image
image
image


alexTerraformation:
Super cool!

src/components/TextTruncated/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    result = tempResult;
+  }
+
+  return {
+    text: result,
+    numberMore: stringsRemaining,
+  };
+}
+
+export default function TextTruncated(props: Props): JSX.Element {
```
charlie-tf, line 80: I'm not sure that this is a great name for this component. Open to suggestions

karthikbtf, line 80: extract props into a type?

src/components/TextTruncated/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const {
+    stringList,
+    maxChars,
```
karthikbtf, line 83: Ideally the table cell renderer will call this component with a list of strings (maybe a ref to the table cell ?).
Component figures out how many characters can be rendered within the width.
Either that or we would need a helper component that can translate available width into maxChars etc here.
Bit of a catch-22 here since table cell width is determined by content displayed.. and here we want to display content by table cell width. A once popular technique is to render hidden elements outside the viewport to do some calculation. Perhaps @constanzauanini might have some suggestions.

karthikbtf (commented):
Thanks for taking this on @charlie-tf , `TruncatedText` is fine .. another option is to look at the objective - display list of strings.. (maybe if that gives some ideas.. i keep coming up with java class names .. i won't suggest anything at this point).
Curious, do we know if there's any npm package that provides similar functionality?
Looks great, have a few suggestions in there.

src/components/TextTruncated/index.tsx
charlie-tf: Do you mean doing 
``` typescript
export default function TextTruncated({prop1, prop2, prop3}: Props): JSX.Element {
```
instead?

src/components/TextTruncated/index.tsx
constanzauanini: What I saw in some libraries is that the component accepts the number of lines to show (https://www.npmjs.com/package/react-truncate) and also the width. We could set fixed width for these cases when we need to truncate? I saw in this library that is has know issues when resizing 😕

charlie-tf:
> do we know if there's any npm package that provides similar functionality?

@karthikbtf I did look around a bit, but the libraries I found worked with a string instead of a list of strings, where based on the ticket we don't want to break in the middle of a word (I implemented an exception for when the first word is really long)

src/components/TextTruncated/index.tsx
charlie-tf: I do [define and export them here](https://github.com/terraware/web-components/pull/107/files#diff-fc13afc62e6095de3b24d5ee26ace1408c6165ab043f7c7d73625b7ccd8954ceR22). I'm sorry for my confusion not trying to be dense 😅

src/components/TextTruncated/index.tsx
karthikbtf: A suggestion, let me know your thoughts.

For initial version:
1) keep component as-is
2) have client pass in a width (as per Constanza's suggestion)
3) use a helper within web-components to map width to maxChars

Once we get this working, we can see if we still need to eliminate 2)..

charlie-tf: okay, that seems like a reasonable approach. Thank you both for your feedback and suggestions!

src/components/TextTruncated/index.tsx
karthikbtf: gripe, my bad.. not sure where my confusion came from.. we don't have a codestyle guideline yet, but.. relocate interface to just above the component function?

charlie-tf: can do!

src/components/TextTruncated/index.tsx
```diff
@@ -77,19 +68,55 @@ function computeFromStringList(
...
+  const pixelsPerChar = useMemo(() => {
+    // compute the pixels per character, averaged over the comma-separated joined stringList
+    const fullText = stringList.join(', ');
+    let result = 0;
+    if (canvasContext) {
+      const fontSize = (textStyle && textStyle.fontSize) || theme.typography.fontSize;
+      const fontFamily = (textStyle && textStyle.fontFamily) || theme.typography.fontFamily;
+
+      canvasContext.font = `${fontSize}px ${fontFamily}`;
+      result = canvasContext.measureText(fullText).width;
```
karthikbtf, line 108: nice 🎉

karthikbtf (approved):
Lets bump the version and give this a whirl in terraware-web, see https://github.com/terraware/terraware-web/blob/main/src/components/Inventory/InventoryCellRenderer.tsx#L35.
Thanks @charlie-tf

PR #114, opened 2022-10-27
Increase version to 1.0.51

charlie-tf:
Just like it says, bump the package.json version to 1.0.51.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.50",
+  "version": "1.0.51",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {


charlie-tf:
the reason is that I increased the version to 1.0.50 for https://github.com/terraware/web-components/pull/107, but then, when merging main, found it already was at 1.0.50, and didn't catch that before merging :/

karthikbtf:
> the reason is that I increased the version to 1.0.50 for #107, but then, when merging main, found it already was at 1.0.50, and didn't catch that before merging :/

oh well, we need to enhance the git action flow with some intelligence.. maybe a checkbox in the PR that says bump version... and the git action takes care of the rest..

karthikbtf (approved):

charlie-tf:
> > the reason is that I increased the version to 1.0.50 for #107, but then, when merging main, found it already was at 1.0.50, and didn't catch that before merging :/
> 
> oh well, we need to enhance the git action flow with some intelligence.. maybe a checkbox in the PR that says bump version... and the git action takes care of the rest..

I'd love to look into how to do this 👍

karthikbtf:
> > > the reason is that I increased the version to 1.0.50 for #107, but then, when merging main, found it already was at 1.0.50, and didn't catch that before merging :/
> > 
> > 
> > oh well, we need to enhance the git action flow with some intelligence.. maybe a checkbox in the PR that says bump version... and the git action takes care of the rest..
> 
> I'd love to look into how to do this 👍

awesome! when things lighten up a bit we can go over this for sure.

PR #116, opened 2022-10-27
TextTruncated import React

charlie-tf:
one more thing I missed... sorry, this problem didn't show up in storybook so I missed it

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.52",
+  "version": "1.0.53",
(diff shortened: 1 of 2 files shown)


constanzauanini (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.52",
+  "version": "1.0.53",
```
karthikbtf, line 3: I missed documenting this (will do that shortly) - storybook changes are not published to npm, version bump here won't capture any new changes in the distribution.

PR #118, opened 2022-10-31
SW-1898 Swapping tokens with more meaningful ones, adjusting some CSS…

chudo76:
Swapping tokens with more meaningful ones, adjusting some CSS and HTML markup.

### DIFF
--- a/src/components/Message/styles.scss
+++ b/src/components/Message/styles.scss
@@ -9,7 +9,7 @@
 
 .tw-message {
   &.tw-message--page {
-    width: 584px;
+    width: 100%;
     &.tw-message--container {
       margin: $tw-spc-base-small 0;
     }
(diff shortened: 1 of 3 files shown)


chudo76:
Accidentally linked this to the wrong Jira.
Should be linked to https://terraformation.atlassian.net/browse/SW-2058

src/components/Message/styles.scss
```diff
@@ -9,7 +9,7 @@
 
 .tw-message {
   &.tw-message--page {
-    width: 584px;
+    width: 100%;
```
karthikbtf, line 12: this might break the web-app, tried 100% in chrome inspector and saw this

Terraware App 2022-10-31 13-55-58

karthikbtf (commented):
might need to bump the package.json version (we will have some intelligence around this soon, Charlie has offered to make it spiffy :) )

src/components/Message/styles.scss
chudo76: I was anticipating Hannah's changes to Page Message (https://www.figma.com/file/aLIG1C6uIMyBWiTopSluxe/Visual-Refresh?node-id=880%3A339230) and made it ahead of time. Hopefully won't be an issue...

karthikbtf: ah nice, i see it now.

karthikbtf (approved):
looks good, @chudo76 - can you bump the version here https://github.com/terraware/web-components/blob/main/package.json#L3 to 54 before we merge this in?

PR #120, opened 2022-11-01
SW-2040 Update passive icon-only button styling

alexTerraformation:
Forgot to include passive and secondary icon-only button styling updates with my last PR.  This should fix all icon-only buttons.
Screen Shot 2022-11-01 at 9 55 07 AM
Screen Shot 2022-11-01 at 9 55 13 AM
Screen Shot 2022-11-01 at 9 55 31 AM
Screen Shot 2022-11-01 at 9 55 39 AM
Screen Shot 2022-11-01 at 9 55 44 AM
Screen Shot 2022-11-01 at 9 55 50 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.54",
+  "version": "1.0.55",
   "author": "Terraformation Inc.",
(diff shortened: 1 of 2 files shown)


karthikbtf (approved):
🎉

PR #122, opened 2022-11-02
SW-2081 Updating icon default size, adjust tooltip, button, select, a…

chudo76:
Updating icon default size, adjust tooltip, button, select, and text field for change.


karthikbtf (approved):
bump package.json version

karthikbtf (approved):

PR #124, opened 2022-11-03
SW-2093 Select row only when clicking checkbox

constanzauanini:
Table already had an option to make the row clickable or not. The problem was that if the row was not clickable, the checkbox didn't worked.

With this change if the row is clickable it will continue working as it used to, but if the row is not clickable only clicking on the checkbox will work.

Already tested with terraware-web

### DIFF
--- a/src/components/table/index.tsx
+++ b/src/components/table/index.tsx
@@ -252,7 +252,11 @@ export default function EnhancedTable<T>({
                         >
                           {showCheckbox && (
                             <TableCell padding='checkbox'>
-                              <Checkbox color='primary' checked={isItemSelected} />
+                              <Checkbox
+                                color='primary'
+                                checked={isItemSelected}
+                                onClick={(e) => (!isClickable || !isClickable(row as T) ? handleClick(e, row as T) : null)}
+                              />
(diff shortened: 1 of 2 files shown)


src/components/table/index.tsx
```diff
@@ -252,7 +252,11 @@ export default function EnhancedTable<T>({
                         >
                           {showCheckbox && (
                             <TableCell padding='checkbox'>
-                              <Checkbox color='primary' checked={isItemSelected} />
+                              <Checkbox
+                                color='primary'
+                                checked={isItemSelected}
+                                onClick={(e) => (!isClickable || !isClickable(row as T) ? handleClick(e, row as T) : null)}
```
karthikbtf, line 258: how is the click handled if the row is clickable? just curious.. the code is a bit confusing that we handleClick only if it is not-clickable.

karthikbtf (approved):
When we pull this into terraware-web, we should remove all the stop propagation fixes?

constanzauanini:
> When we pull this into terraware-web, we should remove all the stop propagation fixes?

Yes, I will do that

PR #125, opened 2022-11-03
SW-2119 Select does not default to readonly

karthikbtf:
- readonly needs to be explicitly set

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.59",
+  "version": "1.0.60",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


constanzauanini:
@karthikbtf I think I did this because if it is not ready only, then you could be able to write (change/edit) the option selected of the dropdown, not only select an option. Readonly is not the same as disabled

karthikbtf:
> @karthikbtf I think I did this because if it is not ready only, then you could be able to write (change/edit) the option selected of the dropdown, not only select an option. Readonly is not the same as disabled

Ah ok, that's good context. So we need to look into @chudo76 's fix for css.

karthikbtf:
Wrong fix, this is not a bug.

PR #126, opened 2022-11-03
SW-2119 Fix readonly vs disabled bgcolor style on select component

karthikbtf:
- move disabled bg color from readonly to disabled

### DIFF
--- a/src/components/Select/styles.scss
+++ b/src/components/Select/styles.scss
@@ -101,6 +101,7 @@
     }
 
     &--disabled {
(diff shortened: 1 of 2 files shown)


src/components/Select/styles.scss
```diff
@@ -281,6 +280,7 @@
       }
 
       &--disabled {
+        background-color: $tw-clr-bg-secondary;
```
constanzauanini, line 283: We need to remove this also

constanzauanini (approved):

PR #127, opened 2022-11-04
Create "allowEditingValue" prop to select

constanzauanini:
- Create new prop "allowEditingValue" to allow writing on selects.
- When select is read-only, don't allow user to change the value (not only by writing but also from selecting a new option)
- I will update terraware-web to use "allowEditingValue"

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -23,7 +23,7 @@ export interface SelectProps {
   onFocus?: () => void;
   fixedMenu?: boolean;
   tooltipTitle?: TooltipProps['title'];
-  allowEditingValue?: boolean;
+  editable?: boolean;
 }
 
 export default function Select(props: SelectProps): JSX.Element {
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -23,6 +23,7 @@ export interface SelectProps {
   onFocus?: () => void;
   fixedMenu?: boolean;
   tooltipTitle?: TooltipProps['title'];
+  allowEditingValue?: boolean;
 }
 
 export default function Select(props: SelectProps): JSX.Element {
--- a/src/components/Select/SelectT.tsx
+++ b/src/components/Select/SelectT.tsx
@@ -30,6 +30,7 @@ export interface SelectTProps<T> {
   toT: (input: string) => T;
   displayLabel: (option: any) => string;
   tooltipTitle?: TooltipProps['title'];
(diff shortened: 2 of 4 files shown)


karthikbtf:
> * Create new prop "allowEditingValue" to allow writing on selects.
> * When select is read-only, don't allow user to change the value (not only by writing but also from selecting a new option)
> * I will update terraware-web to use "allowEditingValue"

What's the difference between readonly and disabled in this component?

src/components/Select/Select.tsx
```diff
@@ -23,6 +23,7 @@ export interface SelectProps {
   onFocus?: () => void;
   fixedMenu?: boolean;
   tooltipTitle?: TooltipProps['title'];
+  allowEditingValue?: boolean;
```
karthikbtf, line 26: probably don't need `Value`.. maybe just `editable` or `allowEdit`?

src/components/Select/SelectT.tsx
```diff
@@ -44,7 +45,7 @@ export default function SelectT<T>(props: SelectTProps<T>): JSX.Element {
     placeholder,
     errorText,
     warningText,
-    readonly = true,
+    readonly,
```
karthikbtf, line 48: What's the difference between readonly and disabled in this component?

karthikbtf (approved):

src/components/Select/SelectT.tsx
constanzauanini: Styles and also readonly allows focus while disabled not

karthikbtf: Do we have a use-case for readonly? This component is rather complex - will be nice to remove any semantics that aren't needed.

karthikbtf: Maybe later

constanzauanini: We don't. Chudo said in slack: "Read-only is used in cases where it may be permission related or maybe one field is read-only because it is dependent on another field's value. We may not need it at all yet in our application."

karthikbtf: Chudo's reference I believe was a bit generic on nomenclature - not sure he was giving us a requirement.
Will be cleaner and less confusing to remove props that aren't needed? I can remove it if we agree it's ok to remove it.

karthikbtf: I can follow up with Chudo, ignore my comment for now.

PR #129, opened 2022-11-04
SW-2168 Add Custom 1400px Breakpoint

charlie-tf:
- set xl breakpoint to 1400px
- this is to support transitioning the background style on the home page
- version 2.0.0 (visual updates)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.62",
+  "version": "2.0.0",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -61,6 +61,15 @@ declare module '@mui/material/styles/createPalette' {
 }
 
 const defaultTheme = {
+  breakpoints: {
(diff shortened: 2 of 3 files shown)


src/theme.ts
```diff
@@ -61,6 +61,15 @@ declare module '@mui/material/styles/createPalette' {
 }
 
 const defaultTheme = {
+  breakpoints: {
+    values: {
+      xs: 0,
+      sm: 600,
+      md: 900,
+      lg: 1200,
+      xl: 1400, // custom xl to transition background on Home page
```
karthikbtf, line 70: Change looks harmless... just curious - any cons to 1400 vs 1536?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "1.0.62",
+  "version": "2.0.0",
```
karthikbtf, line 3: 🎉

karthikbtf (approved):

src/theme.ts
charlie-tf: not that I know of, unless we're breaking any other styles/layouts on `xl` then this shouldn't affect anything

PR #131, opened 2022-11-07
SW-2169 Add Futures Icon

charlie-tf:


karthikbtf (approved):
Is 'Futures' the same as 'Future Forests'?

charlie-tf:
> Is 'Futures' the same as 'Future Forests'?

I'm not sure 🤔 I just used the naming from the asset.

PR #142, opened 2022-11-16
Table Updates

charlie-tf:
- default cell borders
- rm bottom padding of pagination block
- make header hideable

### DIFF
--- a/src/components/table/TableCellRenderer.tsx
+++ b/src/components/table/TableCellRenderer.tsx
@@ -95,6 +95,7 @@ export function CellBooleanRenderer({
   className?: string;
 }): JSX.Element {
   const classes = useStyles();
+
   return (
     <TableCell id={id} align='left' className={classes.default + ' ' + className}>
       <Typography component='p' variant='body1'>
@@ -114,6 +115,7 @@ export function CellNotesRenderer({
   className?: string;
 }): JSX.Element {
   const classes = useStyles();
+  
   return (
     <TableCell id={id} align='left' className={classes.default + ' ' + className}>
       <Typography id={id} component='p' variant='body1'>
--- a/src/components/table/TableCellRenderer.tsx
+++ b/src/components/table/TableCellRenderer.tsx
@@ -115,7 +115,7 @@ export function CellNotesRenderer({
   className?: string;
 }): JSX.Element {
   const classes = useStyles();
-  
+
   return (
     <TableCell id={id} align='left' className={classes.default + ' ' + className}>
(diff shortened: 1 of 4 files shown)


joshbuchea (approved):

src/components/table/TableCellRenderer.tsx
```diff
@@ -54,7 +57,7 @@ export function CellDateRenderer({
   const classes = useStyles();
 
   return (
-    <TableCell id={id} align='left' className={classes.date + ' ' + className}>
+    <TableCell id={id} align='left' className={classes.date + ' ' + classes.default + ' ' + className}>
```
karthikbtf, line 60: another way.. if you think it might be better:
className={`${classes.date}  ${classes.default} ${className}`}

charlie-tf: Oh that's cleaner. I'll make the change 👍

karthikbtf: Thanks 🎉

src/components/table/TableCellRenderer.tsx
```diff
@@ -15,6 +15,9 @@ const useStyles = makeStyles((theme: Theme) => ({
   date: {
     whiteSpace: 'nowrap',
   },
+  default: {
+    borderBottom: `1px solid ${theme.palette.TwClrBrdrSecondary} !important`
```
karthikbtf, line 19: `!important` is usually frowned upon, guessing it's ok here since we don't expect to override this style.

src/components/table/TableHeader.tsx
```diff
@@ -1,10 +1,17 @@
...
 import { SortOrder } from './sort';
 import { TableColumnType } from './types';
 import { SortableContext } from '@dnd-kit/sortable';
 import TableHeaderItem from './TableHeaderItem';
 import { HeadCell } from '.';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  headerCell: {
+    borderBottom: `2px solid ${theme.palette.TwClrBrdrSecondary} !important`,
```
karthikbtf, line 12: we may be able to override the `!important` by looking at what class hierarchy is needed to override the style.
eg.
```
headerCell: {
  '& .somereqdclassname': {
     borderBottom: `2px solid ${theme.palette.TwClrBrdrSecondary}`,
  }
}
```

karthikbtf (commented):
looks mostly good, have some comments re: usage of `! important`

PR #143, opened 2022-11-16
Use More Specific Selector for Table Cells

charlie-tf:
- remove usage of !important


charlie-tf:
wasn't necessary to bump version, so i left it off. can change it if desired 🤷

karthikbtf (approved):
Thanks @charlie-tf 👍

PR #145, opened 2022-11-17
Add font weights to storybook

karthikbtf:
- same solution used in terraware-web
- at some point we may need to converge

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,1 @@
+<link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap" rel="stylesheet" />


.storybook/preview-head.html
```diff
@@ -0,0 +1 @@
+<link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap" rel="stylesheet" />
```
constanzauanini, line 1: So if we are going to use the font form google and not from the local file, we could remove the local file?

karthikbtf: I can remove it. Is this the right solution anyway? We are using google in Terraware web - is that overridden elsewhere in Terraware web?

constanzauanini: I don't think it's overridden. We can merge it like this and if we think i this is the right solution, remove the local file later, because I think there are various sass file referencing the local font

constanzauanini (approved):

joshbuchea:
Thanks for doing this @karthikbtf 🙌

PR #149, opened 2022-11-17
SW-2159 Update Dialog Skrim

charlie-tf:
- skrim off by default
- should be on for destructive actions only
- remove blur
- change background to gradient
image
image

### DIFF
--- a/src/components/DialogBox/DialogBox.tsx
+++ b/src/components/DialogBox/DialogBox.tsx
@@ -15,10 +15,11 @@ export interface Props {
   onClose?: () => void;
   open: boolean;
   scrolled?: boolean;
+  skrim?: boolean;
 }
 
 export default function DialogBox(props: Props): JSX.Element | null {
-  const { title, size, message, children, leftButton, rightButtons, middleButtons, onClose, open, scrolled } = props;
+  const { title, size, message, children, leftButton, rightButtons, middleButtons, onClose, open, scrolled, skrim } = props;
 
   const hasFooter = leftButton || rightButtons || middleButtons;
 
@@ -26,7 +27,7 @@ export default function DialogBox(props: Props): JSX.Element | null {
 
   return open ? (
     <div
-      className={`dialog-box-container ${open ? 'dialog-box--opened' : 'dialog-box--closed'} ${
+      className={`dialog-box-container${skrim ? '--skrim' : ''} ${open ? 'dialog-box--opened' : 'dialog-box--closed'} ${
         isMobile ? 'mobile' : ''
       }`}
     >
--- a/src/components/DialogBox/DialogBox.tsx
+++ b/src/components/DialogBox/DialogBox.tsx
@@ -33,6 +33,7 @@ export default function DialogBox(props: Props): JSX.Element | null {
     >
       <div className={`dialog-box dialog-box--${size}`}>
         <div className='dialog-box--header'>
+          <div className='close-icon-spacer'/>
           <p className='title'>{title}</p>
           <IconButton onClick={onClose} size='small'>
             <Icon name='close' className='icon-close' />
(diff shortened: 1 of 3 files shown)


src/components/DialogBox/styles.scss
```diff
@@ -17,6 +15,17 @@
   z-index: 1300;
   inset: 0px;
 
+  &--skrim {
+    background: linear-gradient(180deg, rgba($tw-clr-base-gray-025, $tw-opcty-base-060), rgba(#EFF5EF, $tw-opcty-base-060));
```
charlie-tf, line 19: Asked for a token for `#EFF5EF`, we didn't think we'd need it at first, but it would be helpful here.

karthikbtf: should we be using semantic tokens here (for the ones with tokens)?
gives designers more control

charlie-tf: was told that base colors should be the way to go for this background gradient

karthikbtf: fair enought

karthikbtf (approved):
👍

src/components/DialogBox/styles.scss
```diff
@@ -17,6 +15,18 @@
   z-index: 1300;
   inset: 0px;
 
+  &--skrim {
+    background-color: rgba($tw-clr-base-gray-025, $tw-opcty-base-060);
+    background-image: linear-gradient(180deg, rgba($tw-clr-base-green-050, $tw-opcty-base-000), rgba($tw-clr-base-green-050, 0.24));
```
charlie-tf, line 20: I was told by @hannah-cai that this was how I should construct the background gradient. Namely, a solid `tw-clr-base-gray-025` overlaid by a `tw-clr-base-green-050` gradient from transparent to 40% (for the skrim, the opacities were multiplied by 0.6 per the spec).

karthikbtf: 👍 
@charlie-tf can we loop @chudo76 in here... using these base tokens is no different from using hardcoded colors, etc.

src/components/DialogBox/DialogBox.tsx
```diff
@@ -15,18 +15,19 @@ export interface Props {
   onClose?: () => void;
   open: boolean;
   scrolled?: boolean;
+  skrim?: boolean;
```
karthikbtf, line 18: oops, missed asking to add this to the stories

karthikbtf: ugh, my bad - it's there. just tried it. all good 👍

karthikbtf (approved):
👍 
my comment re: base tokens.. we can follow up on that later.
would be great to use semantic tokens as much as possible.

src/components/DialogBox/styles.scss
hannah-cai: @chudo76 and I have agreed that not every use of color needs to have a semantic token, we will be using base colors occasionally for things that are more or less one-offs, and which don't have a semantic meaning

karthikbtf: Thanks for clarifying @hannah-cai , sounds good 👍

hannah-cai:
> * skrim off by default
> * should be on for destructive actions only
> * remove blur
> * change background to gradient
> 
> image
> 
> image

unrelated to this branch but it's been bothering me for a while that dialog headers are off-center. is that already in the pipeline to be fixed?

charlie-tf:
>unrelated to this branch but it's been bothering me for a while that dialog headers are off-center. is that already in the pipeline to be fixed?

@hannah-cai I can take a look at it on this branch. I'm not aware of whether it's in the pipeline

charlie-tf:
centered!
image

karthikbtf:
> centered! image

Nice, that's one way to do it, balance it out with a symmetric spacer. 
I was expecting to see a text-align:center and icon floated right or absolutely positioned on the right.
The spacer approach works better with flex for sure. 
All good 🎉

PR #150, opened 2022-11-18
Color Updates

charlie-tf:
- base neutrals were updated with a new ZH link
- some semantic colors are also updated


karthikbtf (approved):
Nice - custom zeroheight link for new theme? 🎉

charlie-tf:
> Nice - custom zeroheight link for new theme? 🎉

Yup! For the base-neutrals @chudo76 gave me a  new link. However, it looks as if some of the semantic colors were updated in the original link as well?

PR #155, opened 2022-11-18
Adjust Cancel Icon Size for TextField

charlie-tf:
- when used as right icon,
  - make it 16x16
  - add some padding-right
- add right icon cancel option to story
image


karthikbtf (approved):
Nice!

PR #157, opened 2022-11-19
Add Transparent Background Option to NavBar

charlie-tf:


src/components/Navbar/styles.scss
```diff
@@ -17,6 +17,10 @@
   padding-right: $tw-sz-base-small;
   padding-top: $tw-sz-base-small;
 
+  &.transparent {
+    background-color: transparent !important;
```
charlie-tf, line 21: made this `!important` because during testing with terraware-web I couldn't figure out how to otherwise make sure it always overrides the assigned background. Seems okay perhaps, since transparency here is a very intentional flag to set?

karthikbtf: agree, this makes sense here.

karthikbtf (approved):
Thanks!

PR #160, opened 2022-11-22
Allow sending a value to onRowClick table function

constanzauanini:


src/components/table/index.tsx
```diff
@@ -45,7 +45,7 @@ export interface Props<T> {
   columns: TableColumnType[];
   rows: T[];
   Renderer?: (props: RendererProps<T>) => JSX.Element;
-  onSelect?: (value: T, fromColumn?: string) => void;
+  onSelect?: (value: T, fromColumn?: string, newValue?: string) => void;
```
karthikbtf, line 48: can we add some comments on how this will work?
looks like we can write a custom cell renderer that calls `onRowClick` with a value.. that's the source of `newValue`

constanzauanini: Yes, sure!

karthikbtf (approved):

PR #163, opened 2022-11-23
Color Updates

charlie-tf:
- json needed fixed to get the correct colors
- this was the source of a button overlay problem


src/style-dictionary-dist2/TerrawareTheme2.ts
```diff
@@ -445,19 +445,19 @@ const TerrawareTheme2: ThemeDictionary = {
     "TwClrBaseWhite": "#FFFFFF",
     "TwClrBg": "#FFFFFF",
     "TwClrBgActive": "#F1F0EC",
-    "TwClrBgGhost": "#FFFFFF",
-    "TwClrBgGhostActive": "#B2AB93",
-    "TwClrBgGhostHover": "#B2AB93",
+    "TwClrBgGhost": "rgba(#FFFFFF, 0)",
+    "TwClrBgGhostActive": "rgba(#B2AB93, 0.4)",
+    "TwClrBgGhostHover": "rgba(#B2AB93, 0.2)",
```
charlie-tf, line 450: @karthikbtf Wondering your thoughts on what we might do here. These theme values will not be usable from TS/JS since it doesn't accept the `rgba(hexvalue, opacity)` format.

karthikbtf: I think I'm missing the problem.
Will this work?
```
<Typography color={theme.palette.TwClrBgGhostHover}>test</Typography>
```
(pardon the hover for color)..

charlie-tf: It doesn't, unfortunately. For example, I changed a button icon fill color like so:
```typescript
  svgIconFill: {
    '& path': {
      fill: theme.palette.TwClrBgGhostHover,
    },
  },
```
and in the browser I see
image

charlie-tf: Rather than saying it "doesn't accept" the format, I should say that it does, but passes it verbatim without transforming it into `rgba(r, g, b, a)` which the browser doesn't seem to like.

charlie-tf: hm, alright, i can see how that might work. i'll play with that some

style-dictionary2/build-theme.mjs
```diff
@@ -1,5 +1,8 @@
+import hexRgb from 'hex-rgb';
+import sd from 'style-dictionary';
```
charlie-tf, line 2: had to change this file to a `.mjs` file and use `import` in order to load the `hexRgb` library.

src/style-dictionary-dist2/TerrawareTheme2.ts
charlie-tf: got it to work! 🎉

karthikbtf: woot!!! will take a look

style-dictionary2/build-theme.mjs
```diff
@@ -55,6 +58,10 @@ export default TerrawareTheme2;`;
         const { name, value } = token;
         let tokenValue = value.value || value;
 
+        tokenValue = tokenValue.replace(/rgba\((#[0-9|a-f|A-F]+),\s*(\d*\.?\d*)\)/g, (match, hex, opacity) => {
```
karthikbtf, line 61: 👍

src/style-dictionary-dist2/terraware.scss
```diff
@@ -87,19 +87,19 @@ $tw-clr-base-transparent: #FFFFFF00;
 $tw-clr-base-white: #FFFFFF;
 $tw-clr-bg: #FFFFFF;
 $tw-clr-bg-active: #F1F0EC;
-$tw-clr-bg-ghost: [object Object];
-$tw-clr-bg-ghost-active: [object Object];
-$tw-clr-bg-ghost-hover: [object Object];
+$tw-clr-bg-ghost: rgba(#FFFFFF, 0);
```
karthikbtf, line 90: do we care about the scss file?

karthikbtf (approved):
one question re: scss, guessing it's ok since it gets compiled later

src/style-dictionary-dist2/terraware.scss
charlie-tf: css seems fine with this format.

karthikbtf: Awesome, thanks!

PR #162, opened 2022-11-22
SW-2124: Responsive Page Layouts -> Empty -> Seed Banks

joshbuchea:
This PR updates the fill color in "blobby icon" SVG assets from blue --> green.

PR #164, opened 2022-11-23
SW-2124: Follow up to update icon color

joshbuchea:
In a [previous PR](https://github.com/terraware/web-components/pull/162) that updated icon fill colors, an incorrect icon was updated. The previous change was reverted and the correct icon fill color has been updated in this PR.


karthikbtf (approved):
👍 thanks for catching this

PR #167, opened 2022-11-28
Upgrade loader utils in web-components to avoid vulnerabilities

karthikbtf:
The dependabot vulnerabilities stem from storybook, upgrading to a patched version.
Tested storybook stories.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,6 @@
   },
   "resolutions": {
     "babel-loader": "8.1.0",
-    "loader-utils": "^2.0.4"
+    "@storybook/react/**/loader-utils": "^2.0.4"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -109,6 +109,7 @@
     "typescript": "^4.3.2"
   },
   "resolutions": {
-    "babel-loader": "8.1.0"
+    "babel-loader": "8.1.0",
+    "@storybook/react/**/loader-utils": "^2.0.4"
```
charlie-tf, line 113: was there no `@storybook/react` version available that upgraded this enough? This is acceptable, but just curious.

charlie-tf (approved):
minor question

package.json
karthikbtf: oh yeah, i went down that path and it was opening up pandora's box. nothing would run and a ton of errors.
we may need to rehaul a lot of our dev dependencies at some point.

charlie-tf: ouch 👍

PR #172, opened 2022-11-30
SW-2157 Icon Updates

charlie-tf:
- `export` icon is deprecated
- `iconExport` should be used instead
- delete `export` icon

Deleted:
image


joshbuchea (approved):
Looks straight forward 👍

PR #178, opened 2022-12-05
Handle error response when npm show package-version returns an error

karthikbtf:
- the previous approach `npm show <version> <property>` was in the right direction as it checks for the specific version directly
- we just need to handle the error, which can be done in the script


karthikbtf:
Would appreciate some comments/thoughts on this.

constanzauanini (approved):

PR #182, opened 2022-12-07
SW-2488 and SW-2487 Decrease size of “x” for mobile nav to match menu icon and fix disabled textfield active style

constanzauanini:
Screenshot 2022-12-07 at 11 24 55

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.27",
+  "version": "2.0.29",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.29",
+  "version": "2.0.28",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.27",
+  "version": "2.0.29",
```
karthikbtf, line 3: should this be 28?

karthikbtf (commented):
should we think about bunching multiple web-components fixes into one version?

src/components/Navbar/Navbar.tsx
```diff
@@ -14,9 +14,9 @@ export interface Props {
 const useStyles = makeStyles(() => ({
   icon: {
     fill: '#708284',
```
karthikbtf, line 16: should this fill be using a token color? (unrelated to your PR but still)

package.json
constanzauanini: I set 28 to the other PR, that's why this is 29

karthikbtf: just saw that, can we merge the 2 fixes in one PR?

src/components/Navbar/Navbar.tsx
karthikbtf: Good call out.. no we need to get new token value from @chudo76 perhaps.
We can follow up separately maybe.

karthikbtf (approved):

src/components/Navbar/Navbar.tsx
chudo76: Token for " x " should be `tw-clr-icn-secondary`. For Kaloko, that would map to `#7F775B`.

constanzauanini: Thanks @chudo76 !

PR #189, opened 2022-12-08
SW-2499 Clicking on read only fields causes the height and padding to change

constanzauanini:


karthikbtf (commented):
@constanzauanini small request, can we make this 14px in this review?
https://github.com/terraware/web-components/blob/main/src/components/TextTruncated/index.tsx#L128

karthikbtf (approved):
👍 thank you!!

src/components/TextTruncated/index.tsx
```diff
@@ -125,7 +125,7 @@ export default function TextTruncated(props: Props): JSX.Element {
             tooltip: classes.tooltip,
           }}
           open={showAllOpen}
-          title={<Typography sx={showAllStyle}>{stringList.join(', ')}</Typography>}
+          title={<Typography sx={{ ...showAllStyle, fontSize: '14px' }}>{stringList.join(', ')}</Typography>}
```
karthikbtf, line 128: showAllStyle can be undefined..
we can do this instead
```
<Typography sx={showAllStyle} fontSize='14px'>
```

karthikbtf (approved):
🎉

PR #191, opened 2022-12-08
SW-2490 Checkbox style fixes

karthikbtf:
- border radius and width are not fixed since it's an icon not css
- added comment in figma for ^
- more fixes to be done in web repo
- check out vercel

Table - Selectable ⋅ Storybook 2022-12-08 08-54-33

Checkbox - Top Align Long Label ⋅ Storybook 2022-12-08 08-54-17

Checkbox - Default ⋅ Storybook 2022-12-08 08-54-01

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.35",
+  "version": "2.0.36",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.34",
+  "version": "2.0.35",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/src/components/Checkbox.tsx
+++ b/src/components/Checkbox.tsx
@@ -67,7 +67,7 @@ export default function Checkbox(props: Props): JSX.Element {
         fontSize: '16px',
         fontWeight: 500,
         lineHeight: '24px',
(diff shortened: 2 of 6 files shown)


karthikbtf:
need to fix disabled status

karthikbtf:
> need to fix disabled status

done with an assumption, will see what @hannah-cai has to say on 0.7 opacity for disabled

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.34",
+  "version": "2.0.35",
```
constanzauanini, line 3: I just merged my PR, so this need to be 36

src/components/Checkbox.tsx
```diff
@@ -1,6 +1,31 @@
...
+  },
+  '&.Mui-checked > .MuiSvgIcon-root': {
+    fill: theme.palette.TwClrBgSelected,
+    color: theme.palette.TwClrBgSelected,
+  },
+  '&.MuiCheckbox-indeterminate > .MuiSvgIcon-root': {
+    fill: theme.palette.TwClrBgSelected,
+    color: theme.palette.TwClrBgSelected,
+  },
+  '&:focus > .MuiSvgIcon-root': {
```
constanzauanini, line 21: It looks like focus has a border radius in figma

karthikbtf: I didn’t add any border radius.. because the inner icon doesn’t have one. Added a comment in figma

constanzauanini (approved):

src/components/Checkbox.tsx
karthikbtf: I added 2 more changes, removed some left margin and disabled animation on table checkboxes.

PR #202, opened 2022-12-14
SW-2513 Move reorder functionality to web components

constanzauanini:
- Reorder of columns was done on accession table instead of table component. Move that behavior to table component.
- Add another column to table story to test reorder there


karthikbtf (commented):
works in vercel 👍 
@constanzauanini bump the version too

karthikbtf (approved):

PR #203, opened 2022-12-15
SW-2537 Fix preselected value on autocomplete

constanzauanini:
Replace use of inputValue with value because:

- inputValue only sets a value for the Input component
- The value prop sends the passed value through getOptionLabel and compares with the options list.

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -66,9 +66,7 @@ export default function Autocomplete({
       {errorText && (
         <div className='textfield-label-container'>
           <Icon name='error' className='textfield-error-text--icon' />
-          <label className='textfield-error-text'>
-            {errorText}
-          </label>
+          <label className='textfield-error-text'>{errorText}</label>
         </div>
       )}
     </div>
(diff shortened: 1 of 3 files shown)


src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -89,14 +87,14 @@ export default function Autocomplete({
       disabled={disabled}
       id={id}
       options={values}
-      getOptionLabel={(option: any) => (option ? (option.label || option) : '')}
+      getOptionLabel={(option: any) => (option ? option.label || option : '')}
       onChange={onChangeHandler}
       onInputChange={(event: any, value: any) => freeSolo && onChangeHandler(event, value)}
-      inputValue={(selected as Option)?.label || selected}
+      value={(selected as Option)?.label || selected}
```
karthikbtf, line 93: Interesting thanks for fixing this.
The selected value highlight (in storybook) is not showing now - probably something I missed.
Can you add this one line 
```
  isEqual: (a: any, b: any) => a.value === b.value,
```
to Complex.args? https://github.com/terraware/web-components/blob/main/src/stories/Autocomplete.stories.tsx#L49

karthikbtf (approved):
Many thanks 🎉

src/components/Autocomplete/Autocomplete.tsx
constanzauanini: Oh, yes, let me try that

src/stories/Autocomplete.stories.tsx
```diff
@@ -49,21 +46,27 @@ Complex.args = {
...
+    {
+      label: 'ma',
+      value: 4,
+    },
+  ],
   onChange: () => true,
   selected: undefined,
   hideClearIcon: true,
   errorText: '',
+  isEqual: (option: any, value) => option.label === value,
```
karthikbtf, line 71: @constanzauanini this is incorrect... the label is not the correct one to check. this will break things downstream.
The option.value is the correct one to check. if we aren't getting an object but just a string,,, this should be changed.

PR #204, opened 2022-12-15
Fix selected value on autocomplete

constanzauanini:


karthikbtf (approved):
Excellent! Thank you!!

PR #206, opened 2022-12-21
SW-2575 Allow toggling open state of nav hiearchy root

karthikbtf:
- we want to allow opening the hierarchy without selecting a menu item

Navbar - Default ⋅ Storybook 2022-12-20 17-02-15

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.50",
+  "version": "2.0.51",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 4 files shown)


constanzauanini:
Why is seeds menu not closing on storybook?

constanzauanini:
> Why is seeds menu not closing on storybook?

Oh, I see. It doesn't have to be selected to allow closing, is that the intended behavior?

karthikbtf:
> Why is seeds menu not closing on storybook?

because it has a selected item - keeping current behavior in staging
we want to open other menus without auto selecting

constanzauanini (approved):

karthikbtf:
> > Why is seeds menu not closing on storybook?
> 
> Oh, I see. It doesn't have to be selected to allow closing, is that the intended behavior?

behavior: allow opening a menu hierarchy without selecting a child menu. on mobile especially this is helpful
the requirements didn't say anything about existing menus that have a child menu selected so not changing that behavior - also if we close that user may be confused.
if they want changes we can iterate

PR #208, opened 2022-12-22
SW-2591 SW-2592 Error and disabled styles for autocomplete

constanzauanini:
Screenshot 2022-12-22 at 09 24 47
Screenshot 2022-12-22 at 09 25 19


karthikbtf (approved):
Thanks for doing this.🎉

PR #210, opened 2022-12-22
SW-2508 Clean up checkbox border radius styles

karthikbtf:
- we were unable to use a border radius before using default mui checkbox
- using additional enhancements to enable this
- see Chudo's findings in the jira https://terraformation.atlassian.net/browse/SW-2508
- the focus outline is still a question on how to do
- feels like maybe the checkbox looks too big now
- test in vercel

Table - Selectable ⋅ Storybook 2022-12-22 14-07-31


karthikbtf:
@hannah-cai let me know if you have any feedback.

src/components/Checkbox.tsx
```diff
@@ -1,27 +1,50 @@
...
+      margin: '2px auto auto',
+    }
   },
-  '&:focus > .MuiSvgIcon-root': {
-    border: `2px solid ${theme.palette.TwClrBgSelected}`,
+  '& > .selected': {
+    border: '2px solid transparent',
+    backgroundColor: theme.palette.TwClrBgSelected,
+    '& > svg' : {
+      fill: 'white',
```
constanzauanini, line 41: Do we have a token for white?

src/components/table/index.tsx
```diff
@@ -321,6 +321,7 @@ export default function EnhancedTable<T>({
                                 color='primary'
                                 checked={isItemSelected}
                                 onClick={(e) => (!isClickable || !isClickable(row as T) ? handleClick(e, row as T) : null)}
+                                {...CheckboxIcons}
```
constanzauanini, line 324: Could we replace the MUI checkbox with the web-component checkbox here so we don't need to pass the icons?

src/components/Checkbox.tsx
karthikbtf: oops, will fix

src/components/table/index.tsx
karthikbtf: This was not working well for some reason. I will try again.

src/components/Checkbox.tsx
karthikbtf: fixed to use theme styles

src/components/table/index.tsx
karthikbtf: Our checkbox is a labeled checkbox.. which has different semantics and event handling.
For now I'm going to keep them this way. We can do a refactor later. It cannot be reused the way it is now.

constanzauanini: Thanks for trying!

constanzauanini (approved):

hannah-cai:
thank you for the updates! the checkboxes look a lot more consistent with everything else now.
a small detail I noticed: the border cuts off the checkmark on the sides.
image
I would also prefer the checkbox to be smaller, but I know that Chudo was pushing for larger checkboxes for bigger click targets. that's something we can still explore though.
I'll leave @chudo76 to be the decider on what we do for focus states, since I'm not as familiar with them.

@karthikbtf a separate issue I noticed is that table rows have a hover state when they're selected and you hover over them. also, clicking on the row selects/de-selects the checkbox; the intended behavior was that you'd need to click on the checkbox itself. these may already be covered in Rachel's table updates, but I wanted to bring them up in case they weren't

karthikbtf:
> thank you for the updates! the checkboxes look a lot more consistent with everything else now. a small detail I noticed: the border cuts off the checkmark on the sides. image I would also prefer the checkbox to be smaller, but I know that Chudo was pushing for larger checkboxes for bigger click targets. that's something we can still explore though. I'll leave @chudo76 to be the decider on what we do for focus states, since I'm not as familiar with them.
> 
> @karthikbtf a separate issue I noticed is that table rows have a hover state when they're selected and you hover over them. also, clicking on the row selects/de-selects the checkbox; the intended behavior was that you'd need to click on the checkbox itself. these may already be covered in Rachel's table updates, but I wanted to bring them up in case they weren't

Thanks @hannah-cai ! I will take a look.

chudo76:
> > thank you for the updates! the checkboxes look a lot more consistent with everything else now. a small detail I noticed: the border cuts off the checkmark on the sides. image I would also prefer the checkbox to be smaller, but I know that Chudo was pushing for larger checkboxes for bigger click targets. that's something we can still explore though. I'll leave @chudo76 to be the decider on what we do for focus states, since I'm not as familiar with them.
> > @karthikbtf a separate issue I noticed is that table rows have a hover state when they're selected and you hover over them. also, clicking on the row selects/de-selects the checkbox; the intended behavior was that you'd need to click on the checkbox itself. these may already be covered in Rachel's table updates, but I wanted to bring them up in case they weren't
> 
> Thanks @hannah-cai ! I will take a look.

Looks like adding `overflow: visible;` to the `svg`'s CSS fixes the clipping issue.

sgrimm:
Closing this since we're using other checkboxes now.

PR #211, opened 2022-12-29
SW-2609 Fix outline cut off on Button component

constanzauanini:
Adding a 5px margin on button element to guarantee the needed space for focus outline


karthikbtf (approved):
looks good, thanks for this change. let's see how it looks in web

PR #212, opened 2022-12-29
Enable version upgrades by dependabot

karthikbtf:
Lets see how this works before we can apply it on terraware-web as well

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "example-name"
+        # For Lodash, ignore all updates
+    open-pull-requests-limit: 15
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
constanzauanini, line 6: Show we do this weekly?

karthikbtf: was thinking we can see sooner how this works.. but yes, can get spammy. will switch to weekly

karthikbtf: Updated this..

karthikbtf:
I'm going to merge this in so we can see something when we get back next week.

constanzauanini (approved):

.github/dependabot.yml
constanzauanini: Thanks!

PR #231, opened 2022-12-30
SW-2631 Separate 'id' and 'propertyName' semantics in various web-components

karthikbtf:
- this allows us to use different ids for tracking purposes
- next step is to apply these changes in terraware-web

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.0.56",
+  "version": "2.0.57",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -13,9 +13,10 @@ export type ValueType = string | Option;
 
 export interface Props {
   id: string;
+  propertyName: string;
   label: string;
   values: ValueType[];
-  onChange: (id: string, value: ValueType) => void;
+  onChange: (propertyName: string, value: ValueType) => void;
   selected: ValueType | undefined;
   freeSolo: boolean;
   disabled?: boolean;
@@ -33,6 +34,7 @@ export type DropdownItem = {
 
 export default function Autocomplete({
   id,
+  propertyName,
   label,
   values,
   onChange,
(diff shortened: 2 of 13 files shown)


constanzauanini:
Can you give an example in terraware-web where we need the id to be different from the propertyName ?

karthikbtf:
> Can you give an example in terraware-web where we need the id to be different from the propertyName ?

Theoretically, everywhere. These are semantically different properties on an element.
One practical example: id='seed-bank-name' propertyName='name' (for editing a seed bank name)
We track by id in google analytics.

karthikbtf:
> > Can you give an example in terraware-web where we need the id to be different from the propertyName ?
> 
> Theoretically, everywhere. These are semantically different properties on an element. One practical example: id='seed-bank-name' propertyName='name' (for editing a seed bank name) We track by id in google analytics.

Another way to do this is to remove 'id' being passed to the onChange callback. We can have the calling code handle it - that may be simpler. This may also be consistent with component frameworks (eg. mui)?
eg.
<component
  id='some-id'
  onChange=(value => setValue('name', value)}
/>

We can maybe discuss this in a team meeting.

karthikbtf:
Updated the ticket with more details.

constanzauanini:
> > > Can you give an example in terraware-web where we need the id to be different from the propertyName ?
> > 
> > 
> > Theoretically, everywhere. These are semantically different properties on an element. One practical example: id='seed-bank-name' propertyName='name' (for editing a seed bank name) We track by id in google analytics.
> 
> Another way to do this is to remove 'id' being passed to the onChange callback. We can have the calling code handle it - that may be simpler. This may also be consistent with component frameworks (eg. mui)? eg. <component id='some-id' onChange=(value => setValue('name', value)} />
> 
> We can maybe discuss this in a team meeting.

Yes, I think this option is simpler!

karthikbtf:
> > > > Can you give an example in terraware-web where we need the id to be different from the propertyName ?
> > > 
> > > 
> > > Theoretically, everywhere. These are semantically different properties on an element. One practical example: id='seed-bank-name' propertyName='name' (for editing a seed bank name) We track by id in google analytics.
> > 
> > 
> > Another way to do this is to remove 'id' being passed to the onChange callback. We can have the calling code handle it - that may be simpler. This may also be consistent with component frameworks (eg. mui)? eg. <component id='some-id' onChange=(value => setValue('name', value)} />
> > We can maybe discuss this in a team meeting.
> 
> Yes, I think this option is simpler!

Thanks, Constanza. I closed this for now. We can discuss it in a meeting maybe, since it's a risky change but essential.

PR #233, opened 2023-01-03
SW-2629 Remove id from on change handlers

constanzauanini:
This change is needed so html id can be set independently of the data model. We can then change the value of the data model using the necessary id, separately of the html id


karthikbtf (commented):
Thanks for doing this @constanzauanini 
I think you need to update the stories as well.
Maybe run storybook just to test.

karthikbtf (approved):

karthikbtf:
oops, i've been merging the other PRs,, i will hold off until this merged

constanzauanini:
> oops, i've been merging the other PRs,, i will hold off until this merged

Merged! Thanks

PR #235, opened 2023-01-04
SW-2561 Rm Dependencies on Old Theme

charlie-tf:
- style-dictionary2 pulls in non-color tokens
- delete refs to old theme scss files
- (bonus) fix color on progress circle


src/components/Navbar/styles.scss
```diff
@@ -26,7 +25,7 @@
   .nav-section {
     padding: $tw-spc-base-x-small $tw-spc-base-small;
     .divider {
-      border-top: $tw-sz-dvdr-stroke solid $tw-clr-brdr-tertiary;
+      border-top: 1px solid $tw-clr-brdr-tertiary;
```
charlie-tf, line 28: The `-divider-` zeroheight files seem to have been removed, likely because that particular component is no longer in use. There are no other `1px` tokens to replace this one with.

karthikbtf (approved):
🎉

PR #237, opened 2023-01-05
Revert "Revert "Bump @mui/styles from 5.9.3 to 5.11.2 (#225)" (#234)"

karthikbtf:
This reverts commit bc60a3b8b6a6a8c1e6de2c0c2383cc714d3431f5.

PR #238, opened 2023-01-05
Emit Type Declarations

charlie-tf:
- removing `emitDeclarationOnly` seems to have broken the type-
  checking in terraware-web
- add back to tsconfig.json
- tested with `yarn ts` in terraware-web locally

image


karthikbtf:
I believe running `yarn test` introduced the previous changes.. might be good to run that just in case.
guessing we need a new distribution / version

charlie-tf:
> I believe running `yarn test` introduced the previous changes.. might be good to run that just in case. guessing we need a new distribution / version

`yarn test` still works. Thanks.

karthikbtf (approved):
You may need to bump the version so we have a new distribution with .d.ts files

charlie-tf:
> You may need to bump the version so we have a new distribution with .d.ts files

@karthikbtf Okay, I'll bump this one to 2.0.61.

You'll want to bump the version in https://github.com/terraware/web-components/pull/237 to 2.0.62 (I still need to figure out that auto-versioning thing 😅)

package.json
```diff
@@ -38,7 +38,7 @@
   "scripts": {
     "start": "start-storybook",
     "build": "build-storybook -c .storybook -o public",
-    "test": "react-scripts test",
+    "test": "cp tsconfig.json tsconfig.json.bak && react-scripts test && cp tsconfig.json.bak tsconfig.json && rm tsconfig.json.bak",
```
karthikbtf, line 41: Optionally use ‘mv’ instead of ‘cp’ and ‘rm’ after the test run

charlie-tf: oh, good call. that saves a command.

karthikbtf:
> > You may need to bump the version so we have a new distribution with .d.ts files
> 
> @karthikbtf Okay, I'll bump this one to 2.0.61.
> 
> You'll want to bump the version in #237 to 2.0.62 (I still need to figure out that auto-versioning thing 😅)

Yes 👍

karthikbtf (approved):

PR #244, opened 2023-01-09
Allow JSX.Elements on Message body

constanzauanini:
This is needed for showing links on the body of the message

### DIFF
--- a/src/components/Message/Message.tsx
+++ b/src/components/Message/Message.tsx
@@ -8,7 +8,7 @@ export interface Props {
   type: 'page' | 'toast';
   priority: 'info' | 'critical' | 'warning' | 'success';
   title?: string | string[];
-  body: string | string[];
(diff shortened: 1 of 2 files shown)


src/components/Message/Message.tsx
```diff
@@ -8,7 +8,7 @@ export interface Props {
   type: 'page' | 'toast';
   priority: 'info' | 'critical' | 'warning' | 'success';
   title?: string | string[];
-  body: string | string[];
+  body: string | (string | JSX.Element)[];
```
karthikbtf, line 11: can we make this 
`string | JSX.Element | (string | JSX.Element)[];`

thanks for doing this

karthikbtf (approved):

src/components/Message/Message.tsx
constanzauanini: Good idea!

PR #246, opened 2023-01-10
SW-2677 Time Zone and Locale for Datepicker

charlie-tf:
The Datepicker component may now accept a time zone and locale as props.

Changing the time zone between e.g. Pacific/Honolulu and Asia/Tokyo, causes the current date to shift, indicating that the component is respecting the change in time zone:
image
image

Changing the locale causes the language of the datepicker to change:
image
image

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -29,9 +29,9 @@ export interface Props {
 
 export default function DatePicker(props: Props): JSX.Element {
   const [temporalValue, setTemporalValue] = useState(props.value || null);
-  const locale = props.locale ?? 'en';
+  const locale = props.locale ?? window.navigator.language ?? 'en';
   React.useEffect(() => {
-    moment.locale([window.navigator.language, locale]);
+    moment.locale([locale]);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [window.navigator.language, locale]);
 
@@ -49,9 +49,8 @@ export default function DatePicker(props: Props): JSX.Element {
     </>
   );
 
-  // set timezone, defaulting to browser
-  const browserTimeZone = Intl.DateTimeFormat(locale).resolvedOptions().timeZone;
-  moment.tz.setDefault(props.defaultTimeZone ?? browserTimeZone);
+  // set timezone, defaulting to 'UTC'
+  moment.tz.setDefault(props.defaultTimeZone ?? 'UTC');
 
   return (
     <div className={`date-picker ${props.className} ${props.errorText ? 'date-picker--error' : ''}`}>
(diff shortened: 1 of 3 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -18,6 +18,8 @@ export interface Props {
   onKeyPress?: KeyboardEventHandler;
   minDate?: any;
   maxDate?: any;
+  defaultTimeZone?: string;
+  locale?: string;
```
karthikbtf, line 22: good thinking!

src/components/DatePicker/DatePicker.tsx
```diff
@@ -46,12 +49,13 @@ export default function DatePicker(props: Props): JSX.Element {
     </>
   );
 
-  // set default timezone to UTC
-  moment.tz.setDefault('UTC');
+  // set timezone, defaulting to browser
+  const browserTimeZone = Intl.DateTimeFormat(locale).resolvedOptions().timeZone;
```
karthikbtf, line 53: We need UTC as default for now otherwise it will break production.

karthikbtf (commented):
looks good, one change needs update

src/components/DatePicker/DatePicker.tsx
karthikbtf: Or we could update all datepickers to pass UTC as default time zone

charlie-tf:
Looks like I still have something missing in the locale aspect of this. The 'es' locale is showing the days of the week as
`M T W T F S S`
While the the 'en' locale it's
`S M T W T F S`
but the weekday labels aren't properly updated. Will investigate further...

src/components/DatePicker/DatePicker.tsx
charlie-tf: I'll go ahead and default to UTC 👍

charlie-tf:
> Looks like I still have something missing in the locale aspect of this. The 'es' locale is showing the days of the week in the calendar as `M T W T F S S` While the the 'en' locale it's `S M T W T F S` but the weekday labels aren't properly updated. Will investigate further...

Oh, it's not only this, but the abbreviations should be entirely different as well. hmm

charlie-tf:
> > Looks like I still have something missing in the locale aspect of this. The 'es' locale is showing the days of the week in the calendar as `M T W T F S S` While the the 'en' locale it's `S M T W T F S` but the weekday labels aren't properly updated. Will investigate further...
> 
> Oh, it's not only this, but the abbreviations should be entirely different as well. hmm

Ah, found the issue in the way the momentjs locale was being set. Fixing...

karthikbtf (approved):
lgtm

PR #248, opened 2023-01-11
SW-2683 Update date utils to support time zones

karthikbtf:
- leveraging luxon
- added tests
- updated with some warnings related fixes


src/utils/date.ts
```diff
@@ -1,13 +1,89 @@
-const getDateDisplayValue = (date: string | number) => new Date(date).toISOString().split('T')[0];
+import { DateTime } from 'luxon';
 
-export const getTodaysDateFormatted = () => {
-  return getDateDisplayValue(Date.now());
+/**
+ * Helper util that returns default time zone as a fallback.
+ */
+const tz = (timeZoneId?: string) => timeZoneId || 'Etc/UTC';
```
charlie-tf, line 6: Would it be useful for this function to validate that `timeZoneId` is a valid identifier?

src/utils/date.ts
```diff
@@ -1,13 +1,89 @@
...
+const getDateFromMillis = (date:any , timeZoneId?: string) => {
+  const millis = new Date(date).getTime();
+
+  return DateTime.fromMillis(millis, { zone: tz(timeZoneId) });
+};
+
+/**
+ * Return the display value for a date in ISO 8601 format.
+ */
+const getDisplayValue = (date: any) => {
```
charlie-tf, line 28: Is `any` the correct type here? Or `Date` or `DateTime`?

src/utils/date.ts
```diff
@@ -1,13 +1,89 @@
...
+/**
+ * Helper util that returns default time zone as a fallback.
+ */
+const tz = (timeZoneId?: string) => timeZoneId || 'Etc/UTC';
+
+/**
+ * Parse a date string with zone implied.
+ * eg. '2023-01-10 in 'America/Los_Angeles'
+ */
+const getDateFromString = (date: any, timeZoneId?: string) => {
```
charlie-tf, line 12: Should this be `date: string`?

charlie-tf: Oh, I see that it is called by `getDate`, which explicitly accepts a `number` as well as a string. Could it be `string | number`?

charlie-tf (approved):
Few questions mostly about type signatures.

src/utils/date.ts
karthikbtf: I thought about it, the concept of valid time zones is now within terraware-web (returned from BE). we can iterate on this for sure.

src/utils/date.ts
karthikbtf: this could be DateTime, let me check

src/utils/date.ts
karthikbtf: also, i realized these could be breaking changes - going to rename this file and restore the original version. We can use the correct file based on opt-in flag.

src/utils/date.ts
karthikbtf: yup DateTime, will update.. and the others are string | number

charlie-tf (approved):
👍

charlie-tf (approved):
:shipit:

PR #249, opened 2023-01-12
SW-2504 left nav hover state updates

karthikbtf:
- selected state color is not available today
- please test in vercel/storybook
- @hannah-cai @chudo76 the $tw-clr-bg-selected color/token is not available today,,, using -active for now
- fixes related to width/padding/scrollbar-width is another PR from terraware-web repo coming soon


src/components/Navbar/styles.scss
```diff
@@ -85,36 +85,17 @@
       padding-left: 20px;
     }
 
-    &--children-selected {
-      background-color: transparent;
```
constanzauanini, line 89: If we remove this, then the "selected parent" will be colored also. Is that intentionally?

karthikbtf: Yes, I got that from the spec.
Navigation 2023-01-13 06-18-03

I don't know if this is correct, will let @hannah-cai comment?

hannah-cai: hmmm. my first instinct is that we shouldn't, but then if users close the dropdown they wouldn't have a way to find which page they're on. I thought that we weren't letting users close dropdowns that they're currently in? currently you can close it.

cc @chudo76 for his thoughts too

karthikbtf: Yes - you can close it now.

hannah-cai: @karthikbtf: @chudo76 and I discussed briefly and we want to keep the nav the way it is currently, where if a dropdown is active you can't close it, and where the highlight is full width even if nav item is a child of a parent. 

on an unrelated note, I'm not sure if this is a bug but the third dropdown's items seem to be incorrect? I'm seeing this happening
image

karthikbtf: the bug is in the test-case, i will clean that up.. and make the other changes.

karthikbtf:
@hannah-cai i made all the requested changes, please take a look in vercel.
there's just one thing that's kinda off.. see image
Navbar - Default ⋅ Storybook 2023-01-17 10-34-07

src/components/Navbar/styles.scss
hannah-cai: @karthikbtf sorry! I'd forgotten to mention this in my earlier comment, but Chudo and I also thought that we should remove the coloring of the parent (in other words, keep this code instead of removing it). if you can't close active dropdowns then users losing their place on the menu is no longer a concern, so we don't need to have the parent highlighted. it also gets rid of the double highlighted issue you pointed out below, and shows more clearly at a glance which tab a user is on

karthikbtf: sounds good 👍

karthikbtf: @hannah-cai do we want the hover state though?

karthikbtf: updated - removed coloring of the parent.

hannah-cai: hmm, good question. is there a way you can turn it off if it's active, but keep it if it isn't?

karthikbtf: Yes, possible, updated with those changes. Please take a look in vercel.

hannah-cai (approved):
looks great! :)

karthikbtf:
> looks great! :)

Awesome, thanks @hannah-cai 👍

PR #250, opened 2023-01-13
Github Action to Auto-Increment Package Version

charlie-tf:
- Comment on PR to apply a patch, minor, or major label to the PR
- Auto-add patch if not a bot
- On PR close, if merged, add a commit to bump the appropriate version

### DIFF
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -1,0 +1,44 @@
+name: Auto Version Based on Label
+
+on:
+  pull_request:
+    branches: [main]
+    types: [closed]
+
+jobs:
+  version-bump:
+    name: 'PR Merged'
+    if: github.event.pull_request.merged == true
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Git config
+        run: |
+          git config --global user.email "${{ github.event.inputs.email }}"
+          git config --global user.name "${{ github.actor }}"
+
+      - name: Apply version bump (major)
+        if: contains(github.event.pull_request.labels.*.name, 'major')
+        run: npm version major
+
+      - name: Apply version bump (minor)
+        if: contains(github.event.pull_request.labels.*.name, 'minor')
+        run: npm version minor
+
+      - name: Apply version bump (patch)
+        if: contains(github.event.pull_request.labels.*.name, 'patch')
+        run: npm version patch
+
+      - name: Commit changes
+        run: |
+          git add -A
+          git commit -m "version bump"
+
+      - name: Git push version bump
+        run: git push origin main --follow-tags --force
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   version-bump:
     name: 'PR Merged'
+    runs-on: ubuntu-latest
     if: github.event.pull_request.merged == true
     steps:
       - name: Checkout main branch
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -29,20 +29,15 @@ jobs:
 
       - name: Apply version bump (major)
         if: contains(github.event.pull_request.labels.*.name, 'major')
-        run: npm version major
+        run: npm version major -m "%s - ${{ github.event.pull_request.title }}"
 
       - name: Apply version bump (minor)
         if: contains(github.event.pull_request.labels.*.name, 'minor')
-        run: npm version minor
+        run: npm version minor -m "%s - ${{ github.event.pull_request.title }}"
 
       - name: Apply version bump (patch)
         if: contains(github.event.pull_request.labels.*.name, 'patch')
-        run: npm version patch
-
-      - name: Commit changes
-        run: |
-          git add -A
-          git commit -m "version bump"
+        run: npm version patch -m "%s - ${{ github.event.pull_request.title }}"
 
       - name: Git push version bump
         run: git push origin main --follow-tags --force
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -10,7 +10,10 @@ jobs:
     name: 'PR Merged'
     if: github.event.pull_request.merged == true
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout main branch
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


karthikbtf:
@charlie-tf This is awesome, how do you add a label to the PR? Or is it looking for any comment with the words 'patch', 'minor', 'major'.etc?

charlie-tf:
> @charlie-tf This is awesome, how do you add a label to the PR? Or is it looking for any comment with the words 'patch', 'minor', 'major'.etc?

You can add it in this "Labels" section on the right hand side:
image

.github/workflows/auto-version.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: npm version minor
+
+      - name: Apply version bump (patch)
+        if: contains(github.event.pull_request.labels.*.name, 'patch')
+        run: npm version patch
+
+      - name: Commit changes
+        run: |
+          git add -A
+          git commit -m "version bump"
```
karthikbtf, line 41: does this happen on the feature-branch prior to merge .. or on the main-branch post-merge?

karthikbtf:
Feature request:
Since now dev won't know which version applies to their change

> > @charlie-tf This is awesome, how do you add a label to the PR? Or is it looking for any comment with the words 'patch', 'minor', 'major'.etc?
> 
> You can add it in this "Labels" section on the right hand side: image

nice! i see it now

.github/workflows/auto-version.yml
charlie-tf: The intent is to run on main-branch post-merge.

charlie-tf: Though if it doesn't look right, please let me know! This is my first time writing one of these 😅

karthikbtf: Nice.
Feature request, not sure if it's possible.. https://docs.npmjs.com/cli/v9/commands/npm-publish
when we publish the new version, wondering if it's possible to label the publish with the PR's title (so devs know which version came from which change). If it's possible, no worries if it's not.
Clearly that would be a separate effort.

charlie-tf:
>Feature request:
Since now dev won't know which version applies to their change

Oh, good point. I should see if I can get it to comment the expected version number or something 🤔

.github/workflows/auto-version.yml
karthikbtf: i believe if we make the bump part of the feature branch - we may run into race conditions if there are parallel PRs being merged.
main branch would be more fool proof since it will bump on top of latest changes.

karthikbtf: basically two PRs merging at the same time may get the same bumped value...
i think main is better for sure.. and bots bump should be ok too maybe.. if we want to pick up a new version based off an upgrade,

.github/workflows/open-pr.yml
```diff
@@ -0,0 +1,35 @@
...
+
+  add-default-labels:
+    name: 'Add Labels'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/github-script@v6
+        # skip auto patch labeling for dependabot or renovate PRs
+        if: ${{ github.actor != 'dependabot[bot]' && github.actor != 'renovate[bot]' }}
```
karthikbtf, line 27: we could enable them.. the public can't merge PRs onto main, we should be ok on that front

karthikbtf: for instance, if you or I or Constanza approves a bot PR and we think it's worth a bump, we can label the PR

.github/workflows/auto-version.yml
charlie-tf: ah, does this mean I should change the checkout to:
```yaml
- uses: actions/checkout@v3
  with:
    ref: main
```
because according to the documentation it will default to checking out the branch of the event that triggered the action, which might be the feature branch if this comes from a PR?

.github/workflows/open-pr.yml
charlie-tf: Oh, you mean allow version bumps from the bot PRs? I thought to disable them because we might not want a version for every single dependency upgrade. But i suppose there's no shortage of integers

karthikbtf: on a semi-related topic, if you think we should use git-commit-sha values instead of npm.. we can discuss that too separately perhaps.

karthikbtf (approved):
looks good to me, we can make changes as we go for sure.. if we need to.
this is awesome, thanks so much!

.github/workflows/open-pr.yml
charlie-tf: >for instance, if you or I or Constanza approves a bot PR and we think it's worth a bump, we can label the PR

That's the workflow I was thinking of—just not applying the label automatically, and making it an intentional action.

karthikbtf: Bots won't cause version bumps unless the PR has a label.. we should be ok?

charlie-tf:
> looks good to me, we can make changes as we go for sure.. if we need to. this is awesome, thanks so much!

Okay, awesome, thanks! I'm going to make that one small change to the `checkout` action ref and then merge. We'll see if it works as expected! 🤞

.github/workflows/auto-version.yml
karthikbtf: ah got it.. if this happens post-merge then yes maybe we should use `main` to avoid competing PRs

karthikbtf: we would have to test it.. with echo/print statements maybe..
there's this https://terraformationteam.slack.com/archives/C01C69KA23Y/p1671532900156529

i used a personal repo to test some of these semantics last time (without actual publishes)

charlie-tf:
Got the `act` tool figured out and did some local testing. Worked out some bugs.

charlie-tf:
Dang, didn't quite work. All the right things happened except the final push, which failed due to branch protection. I'll look into fixing Monday 😞

.github/workflows/auto-version.yml
karthikbtf: I was probably wrong in assuming `/main` won't have race conditions. We could potentially have race conditions on main as well. We would need to, at some point, merge the bump and publish scripts into one,, if publish fails, bump again and retry. For now this is a great.

karthikbtf: Right now we could potentially have two merges bumping to the same version (if they are merged at the same time).
I think only one of them will get published by the publish script (since it won't see any changes in version on npm, when the second bump is committed).

charlie-tf: Ah, I see what you mean. Yeah that would be a good improvement.

PR #256, opened 2023-01-16
Update datepicker value when received value changes

constanzauanini:
Now temporalValue is used to show the datepicker value and that value is initialized with the initial value that comes from props. But if that initial value changes, the datepicker does not updates its temporal value.
Added a useEffect to update temporalValue if 'values' props changes.

Also return selectedValue as a string to consider the timezone

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -33,6 +33,12 @@ export default function DatePicker(props: Props): JSX.Element {
     moment.locale([locale]);
   }, [locale]);
 
+  React.useEffect(() => {
+    if (props.value && props.value !== temporalValue) {
+      setTemporalValue(props.value);
+    }
+  }, [props.value]);
+
   const renderInput = (params: object) => (
     <>
       <TextField {...params} id={props.id} autoFocus={props.autoFocus} onKeyPress={props.onKeyPress} />
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -72,7 +72,7 @@ export default function DatePicker(props: Props): JSX.Element {
           value={temporalValue}
           onChange={(newValue: any) => {
             setTemporalValue(newValue);
-            props.onChange(newValue?.isValid() ? newValue?.toDate() : null);
+            props.onChange(newValue?.isValid() ? newValue?.format('YYYY-MM-DD') : null);
           }}
           renderInput={renderInput}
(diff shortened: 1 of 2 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -33,6 +33,12 @@ export default function DatePicker(props: Props): JSX.Element {
     moment.locale([locale]);
   }, [locale]);
 
+  React.useEffect(() => {
+    if (props.value && props.value !== temporalValue) {
```
karthikbtf, line 37: good catch

src/components/DatePicker/DatePicker.tsx
```diff
@@ -66,7 +72,7 @@ export default function DatePicker(props: Props): JSX.Element {
           value={temporalValue}
           onChange={(newValue: any) => {
             setTemporalValue(newValue);
-            props.onChange(newValue?.isValid() ? newValue?.toDate() : null);
+            props.onChange(newValue?.isValid() ? newValue?.format('YYYY-MM-DD') : null);
```
karthikbtf, line 75: this looks like a breaking change, if client previously handled a JS object, it won't handle this new change. 
thoughts on reverting this change and have client call util to get the formatted value?

constanzauanini: When returning the Date object, the timeZone was lost in the client, that's why I returned the string here. If will try using fromJS in the client and see what happens

charlie-tf:
^removed `patch` since the auto-increment action isn't quite working, and you bumped the version in the PR 🙂

src/components/DatePicker/DatePicker.tsx
karthikbtf: Ok I think I may have missed the objective here. We could possibly return the timeZone as a param in the onChange.
So existing clients that depend on getting back a JS Date in UTC will not be affected.
basically, onChange(jsdate, timeZone)
Thoughts?

constanzauanini: We already have the timeZone in the client, actually the client is passing the timeZone to the datepicker. So I don't think we need to do that. The problem is that when we set the value with getTodaysDateFormatted() we are setting a string, and when the value comes from the onChange it's a date. So I wanted to return a string here to make that cleaner, but I could try creating a new getTodaysDate function that returns a date, and then use .fromJSDate() with the timeZone in the client. What do you think?

karthikbtf (approved):

PR #258, opened 2023-01-17
Auto-version Comment and Label Only on PR Opened

charlie-tf:
Only run when PR is opened to avoid inadvertently adding patch labels.


karthikbtf (approved):
👍

PR #260, opened 2023-01-19
scroll bar bg color

karthikbtf:


src/theme.ts
```diff
@@ -79,6 +79,7 @@ const defaultTheme = {
           WebkitAppearance: 'none',
           width: '4px',
           height: '7px',
+          backgroundColor: '#B2AB9340',
```
karthikbtf, line 82: we can't use themes here - this is where theme is initialized..

charlie-tf (approved):

PR #261, opened 2023-01-20
SW-2733 Support localized content

sgrimm:
Add properties to components that can show text or punctuation to allow (or 
rather, require) callers to pass in appropriately localized values. In cases
where the localized strings need to be rendered dynamically, callers pass in
functions that return the strings.


sgrimm:
This is ready to review, but it's a breaking change so probably best not to merge it until I've published a PR to make the corresponding changes on the terraware-web side.

karthikbtf (approved):
Thanks, looks great 👍

PR #268, opened 2023-01-23
Dropdown wrapper for autocomplete

karthikbtf:
- similar to dropdown for Select

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -16,7 +16,6 @@ export interface Props {
   selected: ValueType | undefined;
   freeSolo: boolean;
   disabled?: boolean;
-  readOnly?: boolean;
   className?: string;
   hideClearIcon?: boolean;
   isEqual?: (option: ValueType, value: ValueType) => void;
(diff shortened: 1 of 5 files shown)


src/components/Dropdown.tsx
```diff
@@ -89,7 +86,40 @@ export default function Dropdown(props: DropdownProps): JSX.Element {
       toT={(str: string) => ({ label: str, value: str } as DropdownItem)}
       displayLabel={(option: DropdownItem) => option?.label || ''}
       onChange={(option: DropdownItem) => onChange(option.value)}
-      tooltipTitle={props.tooltipTitle}
```
karthikbtf, line 92: this is already captured in `remainingProps`

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -91,6 +84,7 @@ export default function Autocomplete({
   return (
     <MUIAutocomplete
       disabled={disabled}
+      readOnly={readOnly}
```
constanzauanini, line 87: Does this work? I tested on story book and it doesn't make any change

karthikbtf: yes i noticed it doesn't seem to do anything.. it's in the API though https://mui.com/material-ui/api/autocomplete/
i guess i can remove it

constanzauanini (approved):

PR #269, opened 2023-01-23
Refactor Dropdown components 

karthikbtf:
- use single component
- use 'autocomplete' boolean option to switch behavior

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -9,12 +9,12 @@ import './styles.scss';
 export type ValueType = string | Option;
 
 export interface Props {
-  id: string;
-  label: string;
-  values: ValueType[];
   onChange: (value: ValueType) => void;
-  selected: ValueType | undefined;
-  freeSolo: boolean;
+  options: ValueType[];
+  id?: string;
+  label?: string;
+  selected?: ValueType | undefined;
+  freeSolo?: boolean;
   disabled?: boolean;
   className?: string;
   hideClearIcon?: boolean;
@@ -27,7 +27,7 @@ export interface Props {
 export default function Autocomplete({
   id,
   label,
-  values,
+  options,
   onChange,
   selected,
   freeSolo,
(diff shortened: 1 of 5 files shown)


constanzauanini (approved):
Looks good to me! just don't understand why the other way (with the component and no the prop) didn't work

karthikbtf:
> Looks good to me! just don't understand why the other way (with the component and no the prop) didn't work

This refactor is mostly to avoid having 2 components. And rename values to options in Autocomplete (to be consistent with options in Select.. and mui calls it options as well).

I'm still running into issues but could be my environment.
When I copy this code to terraware-web it works.. when I copy the web-components dist to my local env it doesn't.

constanzauanini:
> > Looks good to me! just don't understand why the other way (with the component and no the prop) didn't work
> 
> This refactor is mostly to avoid having 2 components. And rename values to options in Autocomplete (to be consistent with options in Select.. and mui calls it options as well).
> 
> I'm still running into issues but could be my environment. When I copy this code to terraware-web it works.. when I copy the web-components dist to my local env it doesn't.

How do you copy this code to terraware-web?

karthikbtf:
> > > Looks good to me! just don't understand why the other way (with the component and no the prop) didn't work
> > 
> > 
> > This refactor is mostly to avoid having 2 components. And rename values to options in Autocomplete (to be consistent with options in Select.. and mui calls it options as well).
> > I'm still running into issues but could be my environment. When I copy this code to terraware-web it works.. when I copy the web-components dist to my local env it doesn't.
> 
> How do you copy this code to terraware-web?

copy over dropdown component file and import from there (with some changes to import paths)..

karthikbtf:
> > > > Looks good to me! just don't understand why the other way (with the component and no the prop) didn't work
> > > 
> > > 
> > > This refactor is mostly to avoid having 2 components. And rename values to options in Autocomplete (to be consistent with options in Select.. and mui calls it options as well).
> > > I'm still running into issues but could be my environment. When I copy this code to terraware-web it works.. when I copy the web-components dist to my local env it doesn't.
> > 
> > 
> > How do you copy this code to terraware-web?
> 
> copy over dropdown component file and import from there (with some changes to import paths)..

Error was in client code.

PR #271, opened 2023-01-24
MultiSelect Fixes

charlie-tf:
- `align-items: flex-start` to position the chevron correctly
- change IconButton to a div to avoid capturing focus and then blurring
- rm padding in pill when there is no label

### DIFF
--- a/src/components/MultiSelect/index.tsx
+++ b/src/components/MultiSelect/index.tsx
@@ -85,9 +85,9 @@ export default function MultiSelect<K, V>(props: MultiSelectProps<K, V>): JSX.El
               />)
             : (<p className='multi-select__placeholder-text' >{placeHolder}</p>)
           }
-          <IconButton className={'multi-select__values__icon-button'} aria-label='show-options'>
-              <Icon name={openedOptions ? 'chevronUp' : 'chevronDown'} className='multi-select__values__icon-right' />
-          </IconButton>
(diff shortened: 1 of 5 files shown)


src/components/MultiSelect/index.tsx
```diff
@@ -85,9 +85,9 @@ export default function MultiSelect<K, V>(props: MultiSelectProps<K, V>): JSX.El
               />)
             : (<p className='multi-select__placeholder-text' >{placeHolder}</p>)
           }
-          <IconButton className={'multi-select__values__icon-button'} aria-label='show-options'>
-              <Icon name={openedOptions ? 'chevronUp' : 'chevronDown'} className='multi-select__values__icon-right' />
-          </IconButton>
+          <div className={'multi-select__values__icon-button'} aria-label='show-options'>
+            <Icon name={openedOptions ? 'chevronUp' : 'chevronDown'} className='multi-select__values__icon-right' />
+          </div>
```
charlie-tf, line 90: The IconButton was capturing focus. Then, when an option was selected it blurred, which propagated up and caused the options list to close. A div doesn't capture focus on the other hand.

src/index.ts
```diff
@@ -59,6 +60,7 @@ export {
   Icon,
   IconTooltip,
   Message,
+  MultiSelect,
```
charlie-tf, line 63: forgot to export this

karthikbtf (approved):
👍

PR #273, opened 2023-01-27
Fix TextTruncated to Accept an Empty List of Strings

charlie-tf:
- optionally render a placeHolder in case of empty stringList
- correctly use the listSeparator

### DIFF
--- a/src/components/TextTruncated/index.tsx
+++ b/src/components/TextTruncated/index.tsx
@@ -38,7 +38,10 @@ interface TextDisplay {
  * @param strings list of strings to process
  * @param maxLength max displayable length
  */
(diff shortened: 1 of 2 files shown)


karthikbtf (approved):
looks good, i would maybe add a story as well.. 
i was able to delete the array elements but unable to set the placeHolder object in storybook...

charlie-tf:
> looks good, i would maybe add a story as well.. i was able to delete the array elements but unable to set the placeHolder object in storybook...

Good call, I will make a story that includes that. Thanks!

karthikbtf (approved):
Nice, thanks!

PR #280, opened 2023-02-02
SW-2712 Initialize theme the same way we do in terraware-web

constanzauanini:
Now the theme is initialized the same way as in terraware-web. 
Still need to update a dependency to check if this works

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,26 +1,24 @@
-import React from 'react'
-import {ThemeProvider} from '@mui/material';
-import {StylesProvider} from '@mui/styles';
-
+import React from 'react';
+import { ThemeProvider, StyledEngineProvider } from '@mui/material';
 
 export const parameters = {
-  actions: { argTypesRegex: "^on[A-Z].*" },
+  actions: { argTypesRegex: '^on[A-Z].*' },
   controls: {
     matchers: {
       color: /(background|color)$/i,
       date: /Date$/,
     },
   },
-}
+};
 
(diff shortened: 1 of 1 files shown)


karthikbtf:
Awesome thanks @constanzauanini - this will help catch issues before we publish new versions that fail in web.

constanzauanini:
> Awesome thanks @constanzauanini - this will help catch issues before we publish new versions that fail in web.

Yes, I still want to test updating a dependency that breaks terraware-web, to see if it breaks here too

constanzauanini:
I just tested updating "@mui/material" from "5.8.7" to "5.11.4" and it breaks here too (not only on terraware-web). So this is working

karthikbtf (approved):
awesome!! now we can try out the mui upgrades next time..
thanks @constanzauanini 🎉

karthikbtf:
> I just tested updating "@mui/material" from "5.8.7" to "5.11.4" and it breaks here too (not only on terraware-web). So this is working

Thanks, sounds good!
btw i removed the patch label since we don't want a version bump for this change.

I think the next step is to figure out how to  upgrade mui successfully.. for later..

constanzauanini:
> > I just tested updating "@mui/material" from "5.8.7" to "5.11.4" and it breaks here too (not only on terraware-web). So this is working
> 
> Thanks, sounds good! btw i removed the patch label since we don't want a version bump for this change.
> 
> I think the next step is to figure out how to upgrade mui successfully.. for later..

Yes, thanks for removing it!

PR #282, opened 2023-02-10
Allow rendering pre-sorted tables 

karthikbtf:


karthikbtf:
using a minor label here

src/components/table/index.tsx
```diff
@@ -176,6 +188,13 @@ export default function EnhancedTable<T extends TableRowType>({
     if (sortHandler) {
       sortHandler(newOrder, property);
     }
+
+    // We want to set the page to 1 on sort order change for presorted tables but we won't have
```
constanzauanini, line 192: remove this maybe?

src/components/table/index.tsx
```diff
@@ -167,6 +169,16 @@ export default function EnhancedTable<T extends TableRowType>({
     }
   }, [rows, setSelectedRows]);
 
+  useEffect(() => {
+    // this is not most elegant but we want to do this if table was sorted by some column in a presorted table
+    // but we don't know when the data changes, hence this useEffect on the data size
+    if (rows.length && isPresorted) {
+      // we want to set page back to 1 if the data changes on presorted lists,
+      // this is because the data was reset due to some sort behavior refetching new data
+      handleChangePage({}, 1);
```
constanzauanini, line 178: Don't we need to execute this for any row change? (Not only a length change)

karthikbtf: ugh good point, the number of rows may stay the same...

karthikbtf: updated, thanks for catching this..

constanzauanini (approved):
Looks good!

PR #285, opened 2023-02-13
Textfield allow 0 value

constanzauanini:
For textfields we were only showing 'truthy values'. Allow when type is number, to have value 0.

### DIFF
--- a/src/components/Textfield/Textfield.tsx
+++ b/src/components/Textfield/Textfield.tsx
@@ -121,7 +121,7 @@ export default function TextField(props: Props): JSX.Element {
             {iconLeft && <Icon name={iconLeft} className='textfield-value--icon-left' />}
             <input
               autoFocus={autoFocus}
-              value={value || ''}
+              value={type === 'number' ? value : value || ''}
               disabled={readonly || disabled}
               placeholder={placeholder}
               onChange={textfieldOnChange}


src/components/Textfield/Textfield.tsx
```diff
@@ -121,7 +121,7 @@ export default function TextField(props: Props): JSX.Element {
             {iconLeft && <Icon name={iconLeft} className='textfield-value--icon-left' />}
             <input
               autoFocus={autoFocus}
-              value={value || ''}
+              value={type === 'number' ? value : value || ''}
```
karthikbtf, line 124: thank you for fixing this,, my bad

karthikbtf (approved):
thanks,, this has been a long standing bug

karthikbtf:
@constanzauanini just one thing.. we don't want to default to showing '0' in some cases (PM preference) .. as long as we can show undefined as empty we should be ok.. i think maybe that's why we had this mixed/strange behavior

constanzauanini:
> @constanzauanini just one thing.. we don't want to default to showing '0' in some cases (PM preference) .. as long as we can show undefined as empty we should be ok.. i think maybe that's why we had this mixed/strange behavior

Oh, you are right. This behavior was for showing empty values. Undefined will work, but I think we have some cases where we need to initialize type "numbers" as zero (not undefined values allowed), and this change, will affect that behavior.

But without this change the seedling details for example are not showing the 0 values where it should.

karthikbtf:
> > @constanzauanini just one thing.. we don't want to default to showing '0' in some cases (PM preference) .. as long as we can show undefined as empty we should be ok.. i think maybe that's why we had this mixed/strange behavior
> 
> Oh, you are right. This behavior was for showing empty values. Undefined will work, but I think we have some cases where we need to initialize type "numbers" as zero (not undefined values allowed), and this change, will affect that behavior.
> 
> But without this change the seedling details for example are not showing the 0 values where it should.

Got it, we can iterate as needed for sure. Thanks for fixing this now.

constanzauanini:
> > @constanzauanini just one thing.. we don't want to default to showing '0' in some cases (PM preference) .. as long as we can show undefined as empty we should be ok.. i think maybe that's why we had this mixed/strange behavior
> 
> Oh, you are right. This behavior was for showing empty values. Undefined will work, but I think we have some cases where we need to initialize type "numbers" as zero (not undefined values allowed), and this change, will affect that behavior.
> 
> But without this change the seedling details for example are not showing the 0 values where it should.

For example when adding a new observation test we don't allow undefined for seedsGerminated, now the textfield type is text so it will not affect the current behavior, but when we update it to type number it will show the zeros. We could create a new type that allows undefined maybe for that case. 

Ok, I will merge this since I think is correct behavior

PR #287, opened 2023-02-16
SW-2899 Fix typing into species field shifting page

constanzauanini:
On selectT component, when a letter is pressed, we get the first element with that letter from the options and scroll it into the view.
In some cases, where we go to the backend to get new options while pressing a letter, the items are reloaded so the scroll into view try to scroll to an item that is not there, ending with a strange position scrolling.

Adding behavior smooth, gives the scroll more time to scroll to the correct position


karthikbtf (approved):
nice, how did you even find the issue..? btw rebase the branch

PR #298, opened 2023-03-03
SW-2530 Create generic tooltip component

alexTerraformation:
To be added to icon buttons.

For the future we could consider replacing the IconTooltip with this generic version but I didn't do that here.

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -27,8 +27,10 @@ export default function Tooltip({ placement = 'top-start', title, children }: To
     setOpen(false);
   };
 
-  const handleTooltipOpen = () => {
-    setOpen(true);
+  const handleTooltipOpen = (event: React.SyntheticEvent | Event) => {
+    if(event.type == 'click' || (event.type === 'mouseover' && !isMobile)){
+      setOpen(true);
+    }
   };
 
   return (
@@ -38,8 +40,9 @@ export default function Tooltip({ placement = 'top-start', title, children }: To
         arrow: classes.arrow,
         tooltip: classes.tooltip,
       }}
+      onOpen={handleTooltipOpen}
       onClose={handleTooltipClose}
-      open={isMobile ? open : undefined}
+      open={open}
       placement={placement}
       sx={{
         maxWidth: isMobile ? '342px' : '464px',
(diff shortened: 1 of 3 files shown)


constanzauanini (approved):
Lgtm! Yes, we should re-use it in icon tooltip later because they have the same code/styles

charlie-tf (commented):
Looks good! The hover interaction is a bit unexpected. It makes sense on non-mobile, but for mobile the tooltip appears on click and closes when the mouse cursor is moved away. Not sure how it behaves on an _actual_ mobile with touch though

charlie-tf:
> Looks good! The hover interaction is a bit unexpected. It makes sense on non-mobile, but for mobile the tooltip appears on click and closes when the mouse cursor is moved away. Not sure how it behaves on an _actual_ mobile with touch though

Okay, I tried it on my phone. The tooltip appears on tapping the button, as expected, and disappears when defocusing, also as expected. It goes away on its own after a couple seconds without any interaction as well.

charlie-tf (approved):
Nice clean component code.

alexTerraformation:
> > Looks good! The hover interaction is a bit unexpected. It makes sense on non-mobile, but for mobile the tooltip appears on click and closes when the mouse cursor is moved away. Not sure how it behaves on an _actual_ mobile with touch though
> 
> Okay, I tried it on my phone. The tooltip appears on tapping the button, as expected, and disappears when defocusing, also as expected. It goes away on its own after a couple seconds without any interaction as well. This seems fine 👍

This comment got me thinking a bit and I realized my solution (and the icontooltip) would not handle responsive transitions because it would try to go controlled<->uncontrolled with the component.  You can demo this in storybook with the icontooltip currently.  I updated it so that it is now always controlled.  If this solution works well then I can port to the icontooltip or replace the icontooltip with this one.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [open, setOpen] = React.useState(false);
+  const { isMobile } = useDeviceInfo();
+  const classes = useStyles();
+
+  const handleTooltipClose = () => {
+    setOpen(false);
+  };
+
+  const handleTooltipOpen = (event: React.SyntheticEvent | Event) => {
+    if(event.type === 'click' || (event.type === 'mouseover' && !isMobile)){
```
charlie-tf, line 31: 🎉

charlie-tf (approved):
New commits look good to me. And it works well in vercel

PR #305, opened 2023-03-10
SW-2518 Apply correct border color to fieldset

alexTerraformation:
This was the simplest solution to fix the issue described in SW-2518.  I tried to clean up some of the styles for these components and found quite a tangle that wasn't worth it to work through right now.

### DIFF
--- a/src/components/Autocomplete/styles.scss
+++ b/src/components/Autocomplete/styles.scss
@@ -49,6 +49,9 @@
     height: '21px';
   };
   & .MuiOutlinedInput-root {
+    fieldset {
+      border-color: $tw-clr-brdr-secondary;
+    }
     &:hover fieldset {
       border-color: $tw-clr-brdr-hover;
     }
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete/styles.scss
```diff
@@ -49,6 +49,9 @@
     height: '21px';
   };
   & .MuiOutlinedInput-root {
+    fieldset {
```
karthikbtf, line 52: definitely simpler 🎉 thanks!!
i think we may need a similar change for the error condition.
if it helps, you can play around in storybook (the error text is settable in the ui)

charlie-tf (approved):
👍

constanzauanini (approved):

src/components/Autocomplete/styles.scss
alexTerraformation: Yes I missed the error border on this one, thanks!

karthikbtf: Looks good 👍 thanks!

karthikbtf (approved):
Lgtm!

PR #306, opened 2023-03-10
SW-2529 Standardize table row height

alexTerraformation:
Previously table rows would adjust based on content & padding.  Design spec'ed row height at 52px.


src/components/table/TableCellRenderer.tsx
```diff
@@ -17,6 +17,9 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   default: {
     '&.MuiTableCell-root': {
+      height: '52px',
```
karthikbtf, line 20: @constanzauanini would this affect the table rows with html input elements?
We can certainly override the height in those few use-cases, good to have this standard height.

karthikbtf (approved):
Looks good. We may need to update terraware-web, there are some tables with html input elements in cells (Reassignment flow, etc..)

src/components/table/TableCellRenderer.tsx
constanzauanini: Yes, this will change all the rows. I don't remember what rows have html apart from species problems, which I checked and this height is enough for the icon we show there.
Also I'm thinking, that if the content of the row is long the width will increase, or the content will be hidden. Is that what we expect?

karthikbtf: You added input element support with the batch withdrawal flow:
Terraware App 2023-03-10 12-31-43

Would a min-height property achieve what we a looking for?

constanzauanini: I think so, but not sure if design wants a min-height or a height. Maybe errors on inputs element would not fit with a height of 52px, but the input itself yes

karthikbtf: those table rows with input elements are 77px height (with padding)

constanzauanini: Yes, but without the padding they fit there. Errors don't, but I'm setting the 52px height in the console, and it doesn't get cut, I don't know why the height doesn't make any effect. Maybe we can merge this and override if necessary?

karthikbtf: Yes we should be able to override with css. 
We have 2 use-cases, Reassign table and batch withdrawals table.

alexTerraformation: @rachel-tf-design - when it comes to these html inputs that won't fit in 52px does design have a recommendation for an override value?  The reassign and batch withdrawals aren't likely to have many rows so increasing in size seems reasonable.

rachel-tf-design: When there's a text field column, the vertical padding should be 6px so that the height still stays as 52px. Would that be possible?

alexTerraformation: No.  The issue is that 52 pixels is not enough to fit the input field plus the error text.

alexTerraformation: Here's what 52px height looks like with an error

rachel-tf-design: Oh I see. Let's keep the 6px vertical padding when it's in the error state (64px height), making the row height 76px.

PR #309, opened 2023-03-14
Add hideSeparator prop to Nav section

constanzauanini:


src/components/Navbar/NavSection.tsx
```diff
@@ -3,14 +3,15 @@ import './styles.scss';
 
 export interface Props {
   title?: string;
+  hideSeparator?: boolean;
```
karthikbtf, line 6: call it `separator` and default it to true in the code?

PR #325, opened 2023-03-31
SW-3361 Move page form / bottom bar and spinner with skrim to web-components

karthikbtf:
- for reuse across apps
- removed additionalLeftButtons as it was not supported (only existed in props)
- created stories


constanzauanini (approved):
Great!

PR #337, opened 2023-04-13
SW-3434 Textfield Can Preserve Newlines

charlie-tf:
When in display mode, setting the preserveNewlines prop to true will cause newlines to be displayed. This is off by default, preserving the previous behavior of collapsing newlines.

image
image

### DIFF
--- a/src/stories/TextfieldNew.stories.tsx
+++ b/src/stories/TextfieldNew.stories.tsx
@@ -67,6 +67,7 @@ Default.args = {
   display: false,
   type: 'text',
   autoFocus: false,
+  preserveNewlines: false,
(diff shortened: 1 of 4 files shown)


karthikbtf (approved):
looks good, one comment on a story

src/stories/TextfieldNew.stories.tsx
```diff
@@ -94,6 +96,7 @@ TextArea.args = {
   display: false,
   type: 'textarea',
   autoFocus: false,
+  preserveNewlines: false,
```
karthikbtf, line 99: set to true?

PR #346, opened 2023-04-27
Add IconMarker

charlie-tf:
generate-assets touched all the icons for some reason

image
image


charlie-tf:
> Looks like the ordering of properties was changed in the generate assets script. Oh well.

Yup

PR #348, opened 2023-04-27
SW-3503 Fix cannot sort table columns that show dates

constanzauanini:
The problem was with using parseFloat. It was returning the value up to the dash character -, ignoring the characters following it. So it was basically comparing the year, if the dates were all same year it didn't change anything when sorting (it did "sort" if you had dates from different years)

Change to use it Number, so that if it cannot compare values as a number, it will continue the logic to compare as a string.

Added a number column and a date column to the stories to test the change


charlie-tf (approved):
Good find! Works well in Vercel

PR #349, opened 2023-04-28
SW-3507 Add SelectPhotos component to web-components

constanzauanini:
Screenshot 2023-04-28 at 12 13 17

### DIFF
--- a/src/components/PhotoChooser.tsx
+++ b/src/components/PhotoChooser.tsx
@@ -42,7 +42,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-export type SelectPhotosProps = {
+export type PhotoChooserProps = {
   title?: string;
   description?: string | string[];
   onPhotosChanged: (photos: File[]) => void;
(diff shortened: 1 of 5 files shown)


karthikbtf (commented):
Export this from index.ts
Also,, thoughs on renaming this to PhotoChooser (like FileChooser).. I called it SelectPhotos initially without much thought.
Maybe we could at some point make this a FileChooser and pass the accepted types as props.. (right now we can't since it has specific photo related iconography in the drag-n-drop pad.. we'd have to customize those too). Just thinking aloud.

constanzauanini:
> ndex.ts

Yes, I think we can adapt this to accept any file in the future if we need it

karthikbtf (approved):

src/components/PhotoChooser.tsx
```diff
@@ -0,0 +1,229 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { Box, Typography, useTheme } from '@mui/material';
+import { Theme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
+import { useDeviceInfo } from '../utils';
+import ErrorBox from './ErrorBox/ErrorBox';
+import Button from './Button/Button';
+import Icon from './Icon/Icon';
```
karthikbtf, line 8: @constanzauanini I don't know how I missed this or how this even builds.. 
let's redo this please

karthikbtf: oops,, wrong place comment

karthikbtf: please ignore

PR #351, opened 2023-04-28
Skip vercel previews for gh-pages

karthikbtf:
- these error out and have no ui code to preview anyways


karthikbtf:
> 👍

Thanks @charlie-tf i went by documentation.. not sure if this will work but let's see

karthikbtf:
ok that didn't work.. sad

PR #363, opened 2023-05-03
Fixes for ViewPhotosDialog

charlie-tf:
- Fix initialSelectedSlide updates
- Fix empty photos list bug
- Make dialog box scroll on overflow (should be good in general I think)

### DIFF
--- a/src/components/ViewPhotosDialog/index.tsx
+++ b/src/components/ViewPhotosDialog/index.tsx
@@ -66,10 +66,14 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
   };
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/components/ViewPhotosDialog/index.tsx
```diff
@@ -121,7 +125,7 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
             {`${selectedSlide+1}/${photos.length}`}
           </Typography>
         ) : undefined}
-        {photos[selectedSlide].decoration}
+        {photos[selectedSlide] && photos[selectedSlide].decoration}
```
charlie-tf, line 128: Here was the bug on empty photos list

PR #364, opened 2023-05-03
Use DialogBox Scrolled Prop

charlie-tf:
The DialogBox component had the `scrolled` prop, which I had missed
earlier. Use that for ViewPhotosDialog instead of making all dialog
boxes scrolled.

### DIFF
--- a/src/components/DialogBox/styles.scss
+++ b/src/components/DialogBox/styles.scss
@@ -118,7 +118,6 @@
     border-bottom: 1px solid $tw-clr-brdr-tertiary;
     padding: $tw-spc-base-medium;
     text-align: center;
-    overflow: scroll;
     &.scrolled {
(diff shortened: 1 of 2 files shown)


src/components/DialogBox/styles.scss
```diff
@@ -118,7 +118,6 @@
     border-bottom: 1px solid $tw-clr-brdr-tertiary;
     padding: $tw-spc-base-medium;
     text-align: center;
-    overflow: scroll;
```
karthikbtf, line 121: Oops, missed this earlier change.. thanks for catching. Scrolled=false allows us to show dialog boxes with dropdowns that don't extend into the dialog body, instead extend outside of the dialog.

karthikbtf (approved):

PR #365, opened 2023-05-04
Update Button Enabled/Disabled on Open

charlie-tf:
The prev/next buttons' disabled state was not being updated on first load. This is exemplified by the case where only one photo was shown. 

This is fixed by adding a call to handleChange in the [initialSelectedSlide, open] useEffect. Thus, it should only be called additionally when the dialog is first opened.

image


constanzauanini (approved):
Thanks!!

PR #367, opened 2023-05-09
Add Right Icon to Buttons

charlie-tf:
image


karthikbtf (approved):
Thanks @charlie-tf .  Right thing to do 🙌 
I think maybe we used text < > earlier because web-components didn't have this right-icon.. and we perhaps took a shortcut.

PR #369, opened 2023-05-10
SW-3558 Remove hardcoded fill color from icons (folder, coin in hand, business network)

karthikbtf:
- this allows us to use custom fill colors


chudo76 (approved):

chudo76:
Looking into if we can provide cleaner exports from Figma. Looks like there may be a plugin that removes default Figma fills:
https://www.figma.com/community/plugin/814345141907543603/SVG-Export
(From: https://levimcg.com/blog/more-control-over-figma-svg-exports/)

I will try this next time before I upload to zeroheight.

karthikbtf:
> Looking into if we can provide cleaner exports from Figma. Looks like there may be a plugin that removes default Figma fills: https://www.figma.com/community/plugin/814345141907543603/SVG-Export (From: https://levimcg.com/blog/more-control-over-figma-svg-exports/)
> 
> I will try this next time before I upload to zeroheight.

Awesome, thanks @chudo76

PR #373, opened 2023-05-15
Fix mobile nav menu styles in web-components

karthikbtf:
- the close icon should be on the left, not right
- also remove extra margin for titles that aren't separated by a divider, this no-separator option was added for funder portal but style wasn't adjusted
- does not affect the terraware use-case

Navbar - Default ⋅ Storybook 2023-05-15 14-14-37


chudo76 (approved):

charlie-tf (commented):
Looks good 👍

karthikbtf:
> Looks good 👍 Thanks for adding to stories

Thanks @charlie-tf !

PR #374, opened 2023-05-16
Allow rendering full images in the view photos dialog

karthikbtf:
- removed the height setting, which triggers loading thumbnails instead of full images
- this leaves it to the client to decide which image to view

### DIFF
--- a/src/components/ViewPhotosDialog/index.tsx
+++ b/src/components/ViewPhotosDialog/index.tsx
@@ -135,7 +135,7 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
               <a href={p.url} target='blank'>
                 <img
                   className='view-photos-dialog-image'
-                  src={`${p.url}?maxHeight=400`}
+                  src={p.url}
                   alt={p.alt}
                   onLoad={() => finishLoading(i)}
                 />


src/components/ViewPhotosDialog/index.tsx
```diff
@@ -135,7 +135,7 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
               <a href={p.url} target='blank'>
                 <img
                   className='view-photos-dialog-image'
-                  src={`${p.url}?maxHeight=400`}
+                  src={p.url}
```
karthikbtf, line 138: @charlie-tf you probably added the height for a reason, do you remember why?
I removed this because it forces rendering of thumbnails only - if client wants to render a full image we can't do that. We could have the client set the maxHeight if necessary.
Would it break anything here?

charlie-tf: I don't think it should break anything. Testing in vercel confirms this. This was a copy/paste from terraware-web

charlie-tf (approved):

src/components/ViewPhotosDialog/index.tsx
karthikbtf: Awesome, thanks!!

PR #378, opened 2023-05-23
SW-3633 PopoverMultiSelect Component

charlie-tf:
- enhance the existing Popover with a renderer prop
- keep track of multiple selection
- left, center, and right menu alignment
- callback on selection change

image


src/components/ButtonDropdownSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { IconName } from '../Icon/icons';
+import Button from '../Button/Button';
+import Icon from '../Icon/Icon';
+import './styles.scss';
+
+type ButtonDropdownSelectProps = {
+  id: string;
+  options: DropdownItem[];
+  initialSelection: DropdownItem[];
+  onChange: (selection: DropdownItem[]) => void;
```
karthikbtf, line 13: Values alone should be enough? Dropdown items have more semantics (rendered label etc..)

src/components/ButtonDropdownSelect/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return <>
+    {buttonStyle === 'terraware' ? (
+      <Button onClick={handleClick} label={buttonLabel} icon={buttonIcon} disabled={disabled}/>
+    ) : (
+      <button onClick={handleClick} className='button-dropdown-select__plain-button' disabled={disabled}>
+        {buttonIcon && <Icon name={buttonIcon} size='small' className='plain-button__icon'/>}
+        {buttonLabel && <Typography className='plain-button__label'>{buttonLabel}</Typography>}
+      </button>
+    )}
+    <Menu
```
karthikbtf, line 59: Thoughts on using the existing [popover component](https://github.com/terraware/web-components/blob/main/src/components/PopoverMenu/Popover.tsx)? We could enhance that with a an optional renderer prop (for the label/check, etc.).
If it's not feasible, that's fine.

karthikbtf (commented):
This looks cool!
Just a thought/question - should this be a dedicated layers component in terraware-web - supported by some enhancements to the Popover component in web-components?

src/components/ButtonDropdownSelect/index.tsx
charlie-tf: Okay, I changed this and I like it much more. I kept the component here, feeling it is generic enough for general use. The Popover component was used, with an enhancement for custom item rendering as you suggested.

charlie-tf: I'll create the MapLayerSelect as a component in terraware-web, built on this

karthikbtf: > Okay, I changed this and I like it much more. I kept the component here, feeling it is generic enough for general use. The Popover component was used, with an enhancement for custom item rendering as you suggested.

Thanks for following up @charlie-tf - I will take a look

src/components/PopoverMenu/Popover.tsx
```diff
@@ -76,7 +78,7 @@ export default function PopoverDropdown(props: PopoverProps): JSX.Element {
                   sx={itemStyles(item.type)}
                   disableRipple={true}
                 >
-                  {item.label}
+                  {itemRenderer ? itemRenderer(item) : item.label}
```
karthikbtf, line 81: nice, thanks!

src/components/PopoverMenuMultiSelect/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  anchorElement: HTMLElement | null;
+  initialSelection: any[];
+  menuAlign?: 'left' | 'center' | 'right';
+  onChange: (selection: any[]) => void;
+  options: any[];
+  optionRenderer: (option: any) => string;
+  setAnchorElement: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
+};
+
+export default function PopoverMenuMultiSelect(props: PopoverMenuMultiSelectProps): JSX.Element {
```
karthikbtf, line 18: thanks for the rename, this is more apt for sure

src/components/PopoverMenuMultiSelect/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Popover from '../PopoverMenu/Popover';
+
+export type PopoverMenuMultiSelectProps = {
+  anchorElement: HTMLElement | null;
+  initialSelection: any[];
+  menuAlign?: 'left' | 'center' | 'right';
+  onChange: (selection: any[]) => void;
+  options: any[];
+  optionRenderer: (option: any) => string;
+  setAnchorElement: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
```
karthikbtf, line 15: ok to make this a `(anchorEl: HTMLElement) => void` .. not a huge fan of exposing underling mechanics of how the anchor element is set

src/components/PopoverMenuMultiSelect/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import Icon from '../Icon/Icon';
+import './styles.scss';
+import Popover from '../PopoverMenu/Popover';
+
+export type PopoverMenuMultiSelectProps = {
+  anchorElement: HTMLElement | null;
+  initialSelection: any[];
+  menuAlign?: 'left' | 'center' | 'right';
+  onChange: (selection: any[]) => void;
+  options: any[];
```
karthikbtf, line 13: use DropdownItem prop and only send values back? Same for initial selection. The labels are not required for selection.

src/components/PopoverMenuMultiSelect/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      if (itemIndex >= 0) {
+        removeFromSelection(itemIndex);
+      } else {
+        addToSelection(item.value);
+      }
+    }}
+    anchorElement={anchorElement}
+    setAnchorElement={setAnchorElement}
+    itemRenderer={(item) => {
+      const itemFound = selection.find((s) => s === item.value);
```
karthikbtf, line 54: rename itemFound to selected or something like that?

karthikbtf (commented):
This looks great, thanks for simplifying this
A few suggestions, would like to hear your thoughts and then I can review again.

karthikbtf (approved):
Approving since the suggestions were mostly cosmetic. Functionally good 👍 
Might be good to change the PR title.

charlie-tf:
> Approving since the suggestions were mostly cosmetic. Functionally good 👍 Might be good to change the PR title.

Thanks for the review @karthikbtf !

src/components/PopoverMenuMultiSelect/index.tsx
charlie-tf: I think I'll rename it to `PopoverMultiSelect`. It does not quite line up with the `PopoverMenu` component, which also renders an anchor. Rather, it is a wrapper around `Popover`

src/components/PopoverMenuMultiSelect/index.tsx
charlie-tf: I changed options to `Section[]` to match the `Popover` api, but I left `initialSelection` as an `any[]`, since that matches the type of the `onChange` callback argument.

karthikbtf: That seems like the right thing to do 👍 
Changes look great, thanks for the iterations!! 🎉

PR #381, opened 2023-05-25
Remove DatePicker change on useEffect

karthikbtf:
- Added this yesterday because of a react compile warning
- It is causing issues reported by Alex
- was able to test this locally
- guessing it was left out earlier for the same reason


charlie-tf (approved):
LGTM

PR #382, opened 2023-05-25
SW-3632 Slider Component

charlie-tf:
- Create a Slider component
- Wraps a basic MUI Slider and adds TW styling

image

### DIFF
--- a/src/stories/Slider.stories.tsx
+++ b/src/stories/Slider.stories.tsx
@@ -1,0 +1,57 @@
+import { Story } from '@storybook/react';
+import React from 'react';
+import { Box } from '@mui/material';
+import Slider, { SliderProps } from '../components/Slider';
+
+export default {
+  title: 'Slider',
+  component: Slider,
+};
+
+const Template: Story<SliderProps> = (args) => {
+  return <Box height='100vh' display='flex' alignItems='center'>
+    <Slider {...args} />
+  </Box>;
+};
+
+export const Default = Template.bind({});
+
+export const SnapToMark = Template.bind({});
+
+const marks = [
+  {
+    label: '0%',
+    value: 0,
+  },
+  {
+    label: '25%',
+    value: 25,
+  },
+  {
+    label: '50%',
+    value: 50,
+  },
+  {
+    label: '75%',
+    value: 75,
+  },
+  {
+    label: '100%',
+    value: 100,
+  },
+];
+
+Default.args = {
+  defaultValue: 50,
+  min: 0,
+  max: 100,
+  step: 1,
+  marks,
+  valueLabelDisplay: 'auto',
+};
+
+SnapToMark.args = {
+  defaultValue: 50,
+  marks,
+  valueLabelDisplay: 'auto',
+};
--- a/src/stories/Slider.stories.tsx
+++ b/src/stories/Slider.stories.tsx
@@ -52,5 +52,4 @@ SnapToMark.args = {
   defaultValue: 50,
   marks,
   valueLabelDisplay: 'auto',
-  onChange: (value) => alert(`Power level at ${value}%`)
 };
--- a/src/stories/Slider.stories.tsx
+++ b/src/stories/Slider.stories.tsx
@@ -52,4 +52,5 @@ SnapToMark.args = {
   defaultValue: 50,
   marks,
   valueLabelDisplay: 'auto',
+  onChange: (value) => alert(`Power level at ${value}%`),
 };
--- a/src/stories/Slider.stories.tsx
+++ b/src/stories/Slider.stories.tsx
@@ -54,4 +54,5 @@ SnapToMark.args = {
   defaultValue: 50,
   marks,
   valueLabelDisplay: 'auto',
+  onChange: (value) => alert(`Power level at ${value}%`)
 };
--- a/src/stories/Slider.stories.tsx
+++ b/src/stories/Slider.stories.tsx
@@ -24,7 +24,6 @@ const marks = [
     value: 0,
   },
   {
-    label: '25%',
     value: 25,
   },
   {
@@ -32,7 +31,6 @@ const marks = [
     value: 50,
   },
   {
-    label: '75%',
     value: 75,
   },
   {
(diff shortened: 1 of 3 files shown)


karthikbtf:
This looks nice @charlie-tf 
We don't want intermediate stops (between marks) - can we try to snap it to nearest mark? Wondering if there is an option.

charlie-tf:
>We don't want intermediate stops (between marks) - can we try to snap it to nearest mark? Wondering if there is an option.

@karthikbtf Yup! There's a second story that illustrates that case.

karthikbtf:
> > We don't want intermediate stops (between marks) - can we try to snap it to nearest mark? Wondering if there is an option.
> 
> @karthikbtf Yup! There's a second story that illustrates that case.

Awesome, missed that, can we remove the pop-up? Not able to test the sliding/snap behavior.

src/stories/Slider.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  step: 1,
+  marks,
+  valueLabelDisplay: 'auto',
+};
+
+SnapToMark.args = {
+  defaultValue: 50,
+  marks,
+  valueLabelDisplay: 'auto',
+  onChange: (value) => alert(`Power level at ${value}%`)
```
karthikbtf: not sure why this is kicking in before the drag/snap

charlie-tf:
> > > We don't want intermediate stops (between marks) - can we try to snap it to nearest mark? Wondering if there is an option.
> > 
> > 
> > @karthikbtf Yup! There's a second story that illustrates that case.
> 
> Awesome, missed that, can we remove the pop-up? Not able to test the sliding/snap behavior.

Yes, I can do that

src/stories/Slider.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    label: '100%',
+    value: 100,
+  },
+];
+
+Default.args = {
+  defaultValue: 50,
+  min: 0,
+  max: 100,
+  step: 1,
```
karthikbtf, line 46: does this influence the snap? absence of step = snap to mark?

karthikbtf (approved):
Looks good, a few questions.

src/stories/Slider.stories.tsx
charlie-tf: Yes, it snaps to mark if `step` is missing

src/stories/Slider.stories.tsx
charlie-tf: Yeah, i noticed that, too, and wasn't really thrilled about it. I'll dig a little bit to see whether I can change that behavior

src/stories/Slider.stories.tsx
karthikbtf: Thanks 👍  makes sense - that's clever design by Mui. 
I noticed that it doesn't animate the drag.. guessing that's ok but may want to check with Chudo.

src/stories/Slider.stories.tsx
karthikbtf: ignore this comment.. looks like the drag is not animated, maybe mui behavior

karthikbtf: > Yeah, i noticed that, too, and wasn't really thrilled about it. I'll dig a little bit to see whether I can change that behavior

Sounds good, this seems like a good initial version maybe?

charlie-tf: I found a simple fix: using `onChangeCommitted` instead of `onChange` triggers on mouseup, meaning you can slide the slider around before getting the popup. I'll add the popup back in to that story to illustrate.

karthikbtf: Woot!

charlie-tf: Thanks for the callout here. I had just passed it over, but the solution was simple to find

karthikbtf:
It occurred to me, we could use a step of 1 to show the drag animation.. onChangeCommitted we could set the value to the next nearest mark. In case it helps. Looking at the API, didn't find anything obvious. There's another [solution (complex)](https://stackoverflow.com/questions/60663602/material-ui-slider-wont-slide) on stackoverflow. In case any of these help.

The use-case for us will be the various observation dates (YYYY-MM-DD), will we convert them to timestamps in order to set the mark value?

src/stories/Slider.stories.tsx
karthikbtf: Sure 👍 glad they have a prop/callback. The only other thing is showing the drag before the snap, looks like mui has no out-of-box solution for that. We could explore later if Chudo wants, maybe.

karthikbtf:
> It occurred to me, we could use a step of 1 to show the drag animation.. onChangeCommitted we could set the value to the next nearest mark. Looking at the API, didn't find anything obvious. There's another [solution (complex)](https://stackoverflow.com/questions/60663602/material-ui-slider-wont-slide) on stackoverflow. In case any of these help.
> 
> The use-case for us will be the various observation dates (YYYY-MM-DD), will we convert them to timestamps in order to set the mark value?

Fwiw this ^ should not prevent merging. Looks great as it is now 👍

charlie-tf:
> > It occurred to me, we could use a step of 1 to show the drag animation.. onChangeCommitted we could set the value to the next nearest mark. Looking at the API, didn't find anything obvious. There's another [solution (complex)](https://stackoverflow.com/questions/60663602/material-ui-slider-wont-slide) on stackoverflow. In case any of these help.
> > The use-case for us will be the various observation dates (YYYY-MM-DD), will we convert them to timestamps in order to set the mark value?
> 
> Fwiw this ^ should not prevent merging. Looks great as it is now 👍

Yeah, I'll go ahead and merge this, to get on to working on the date picker version of this, and consider the animation solution later. Thanks so much!

charlie-tf:
>The use-case for us will be the various observation dates (YYYY-MM-DD), will we convert them to timestamps in order to set the mark value?

On this point, yes, I was thinking to convert to timestamps for the slider mark values, which will correctly position them relatively on the slider.

PR #390, opened 2023-06-13
Make Slider Value Controllable

charlie-tf:
Allow the consuming component to set the value of the slider.

### DIFF
--- a/src/components/Slider/index.tsx
+++ b/src/components/Slider/index.tsx
@@ -8,6 +8,7 @@ export type SliderMark = {
 
 export type SliderProps = {
   defaultValue?: number;
+  value?: number;
   marks?: SliderMark[];
   min?: number;
   max?: number;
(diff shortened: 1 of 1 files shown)


src/components/Slider/index.tsx
```diff
@@ -43,18 +44,19 @@ export default function Slider(props: SliderProps): JSX.Element {
...
   const [currentValue, setCurrentValue] = useState(defaultValue);
-  const handleChange = (event: React.SyntheticEvent | Event, value: number | number[]) => {
-    if (!Array.isArray(value)) {
-      if (value !== currentValue) {
-        setCurrentValue(value);
-        onChange(value);
+  const handleChange = (event: React.SyntheticEvent | Event, val: number | number[]) => {
+    if (!Array.isArray(val)) {
+      if (val !== currentValue) {
+        setCurrentValue(val);
```
karthikbtf, line 50: do we need the current value?

charlie-tf: Ah, i neglected that. I can probably get rid of it

karthikbtf (approved):
Looks good!

PR #392, opened 2023-06-13
SW-3757 View photos dialog react to initialSelectedSlide prop

karthikbtf:
- the initialSelectedSlide and open useEffects had been merged, which broke the handling of slide change (the handleChange ignores selected slide change)
- separated out the concerns, which was the right thing to do anyways
- also reorganized the code a bit

### DIFF
--- a/src/components/ViewPhotosDialog/index.tsx
+++ b/src/components/ViewPhotosDialog/index.tsx
@@ -50,7 +50,7 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
     },
   };
 
-  const handleChange = () => {
+  const handleChange = useCallback(() => {
     if (myCarousel.current) {
       if (myCarousel.current.state.currentSlide + 1 >= photos.length) {
         setIsNextDisabled(true);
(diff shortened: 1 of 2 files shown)


karthikbtf:
I believe this is the right fix (tested in storybook by updating initialSelectedSlide). cc @constanzauanini has more context here.

charlie-tf (approved):
Thanks for doing this

src/components/ViewPhotosDialog/index.tsx
```diff
@@ -68,14 +62,20 @@ export default function ViewPhotosDialog(props: ViewPhotosDialogProps): JSX.Elem
       } else {
         setIsPreviousDisabled(false);
       }
-      setSelectedSlide(myCarousel.current.state.currentSlide);
```
karthikbtf, line 71: This was the primary offender for clients that want to open the modal pointing to a specific slide (specific slide was not honored since this line overrides the specific slide).
Also existing code cannot support opening to a specific slide without using a setTimeout (which is fine for now, we can revisit later).

constanzauanini (approved):

PR #395, opened 2023-06-20
SW-3771 Do Not Conditionally Render Styles Hook

charlie-tf:
- The makeStyles function returns a hook, which must not be conditionally called.
- Reorganized the code so that the hook is unconditionally called.

### DIFF
--- a/src/components/table/TableHeaderItem.tsx
+++ b/src/components/table/TableHeaderItem.tsx
@@ -8,6 +8,16 @@ import { makeStyles } from '@mui/styles';
 import { SortOrder } from './sort';
 import IconTooltip from '../IconTooltip';
 
+const useStyles = makeStyles((theme: Theme) => ({
+  dragIcon: {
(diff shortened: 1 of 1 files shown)


karthikbtf (approved):
Wow 👏 this is a great one!! I ran into this locally a few times but was never able to repro consistently. I did see the issue come from tables.

charlie-tf:
> > > Wow 👏 this is a great one!! I ran into this locally a few times but was never able to repro consistently. I did see the issue come from tables.
> > 
> > 
> > Yeah, Becky was hitting this consistently, and her stack trace helped me pinpoint it 
> 
> Wonder why none of us ran into this consistently. I happened to run into this only when I did some manual copy of a locally built web-component distribution into node_modules (but not all the time). There is something non-deterministic in the execution for sure.

Not sure, it was conditional on both `headCell.label` and `i > 0`, and I wonder whether there was a race condition in those conditions being met

PR #396, opened 2023-06-21
SW-3712 add a fillColor property on Icon component

karthikbtf:
- make it easier to fill an icon with color
- avoids having to override fill with nested css
- did not extend Button with similar pattern, different use-case (icon fill is propagated from button)

Icon - Default ⋅ Storybook 2023-06-21 14-47-09

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -1,16 +1,31 @@
+import React from 'react';
+import { makeStyles } from '@mui/styles';
 import { Size } from '../Size';
 import icons, { IconName } from './icons';
 import './styles.scss';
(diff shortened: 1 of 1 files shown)


charlie-tf (approved):
Hey, nice improvement! That makes this component much nicer to use :)

karthikbtf:
> Hey, nice improvement! That makes this component much nicer to use :)

Yes, we can remove all those makeStyles for fill colors :)

karthikbtf:
> > Hey, nice improvement! That makes this component much nicer to use :)
> 
> Yes, we can remove all those makeStyles for fill colors :)

I will create a ticket to replace them with the fill prop

PR #403, opened 2023-07-14
Tabs component in web-components

karthikbtf:
- copied over from funder-web, i believe the styles are the same
- added a story
- next step is to refactor funder-web to use this, which has some additional semantics
- and also use this in terraware-web

Tabs - Default ⋅ Storybook 2023-07-14 12-03-59


charlie-tf (approved):
This looks great! 🎉

karthikbtf:
I missed exporting this :(

PR #408, opened 2023-07-19
Fix table/cell boolean renderer to also consider boolean

karthikbtf:


src/components/table/TableCellRenderer.tsx
```diff
@@ -108,7 +108,7 @@ export function CellBooleanRenderer({
   return (
     <TableCell id={id} align='left' className={`${classes.default} ${className}`}>
       <Typography component='p' variant='body1' fontSize='14px'>
-        {value === 'true' ? booleanTrueText : booleanFalseText}
+        {value?.toString() === 'true' ? booleanTrueText : booleanFalseText}
```
constanzauanini, line 111: Should we also modify line 101 to accept boolean? (Not sure if that would generate an error or not)

constanzauanini (approved):

src/components/table/TableCellRenderer.tsx
karthikbtf: I will change it. Won't impact anything since the code change calls a toString() which exist on all javascript objects/types. Though.. the types on 101 look kinda wrong... numer/any/ReactNode are not applicable. I won't change those since I don't know what will break :)

karthikbtf: I will make it `any`.. just adding `boolean` there does nothing. we don't cast the input data as boolean anywhere. I think we aren't really leveraging typescript correctly in the renderers since we are using the renderer type ('string', 'boolean', 'number' ,etc) to define what kind of data it is.

constanzauanini: Sounds good! Thanks for doing this!

PR #418, opened 2023-08-23
SW-3800 Put blank values last when doing FE sorting

constanzauanini:
Blank values where changing from top to end because we where changing the sign (-) when changing order to the descendingComparator.
Fix is to move the order inside the descendingComparator and not change the result for blank values (always at the end)


src/components/table/sort.ts
```diff
@@ -9,11 +9,20 @@ export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): number {
   const bValue = b[orderBy] ?? '';
   const aValue = a[orderBy] ?? '';
 
-  if (bValue < aValue) {
+  // blank values at the end always (any order)
+  if (aValue === '' || aValue === null) {
```
karthikbtf, line 13: suggestion use trim:
```
  if (aValue === null || aValue === undefined || aValue.trim() === '') {
  }
```

src/components/table/sort.ts
```diff
@@ -9,11 +9,20 @@ export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): number {
...
+    return 1;
+  }
+
+  if (bValue === '' || bValue === null) {
```
karthikbtf, line 17: same, trim

src/components/table/sort.ts
```diff
@@ -9,11 +9,20 @@ export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): number {
...
     return -1;
   }
+
```
karthikbtf, line 20: if aValue and bValue are both null/undefined/'', we should return 0

karthikbtf (commented):
looks good i think, some thoughts

src/components/table/sort.ts
```diff
@@ -9,11 +9,20 @@ export function descendingComparator<T>(a: T, b: T, orderBy: keyof T): number {
...
+  if (bValue < aValue) {
+    return order === 'desc' ? -1 : 1;
```
karthikbtf, line 22: nice, we may need to check 'desc' for aValue/bValue empty state as well and return -1/1 accordingly.

src/components/table/sort.ts
karthikbtf: just fyi - `trim` will only work if it is a string, maybe aValue.toString().trim()?

karthikbtf (approved):
Should be good to merge with the suggestions followed up 👍

src/components/table/sort.ts
karthikbtf: actually nvm, i realize we want this at the bottom, no need to check for `desc`.. my bad

PR #424, opened 2023-09-05
Lock Node version to avoid Storybook errors

nickgraz:
- Lock Node version for local development to 16 via `.nvmrc`
- Add some notes to README


constanzauanini (approved):
Nice!

PR #425, opened 2023-09-05
GH Workflow for publishing NPM RC build

nickgraz:
- Add GH workflow to publish NPM package for RC versions on specific comment
- Specific comment is "publish rc"

- [x] Verify RC packages work as expected for local dev
- [ ] Remove testing action from workflow
- [ ] Add back comment check to workflow


nickgraz:
publish rc

nickgraz:
publish rc

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.3.58",
+  "version": "2.3.59-rc.3",
```
charlie-tf, line 3: Should this merge? Will it be properly overwritten to `2.3.59`?

charlie-tf: The `auto-version.yaml` workflow will run a (in this case) `npm version patch` command

charlie-tf (commented):
Very cool! Just a question about what the version ends up as post-merge.

.github/workflows/publish-rc.yaml
```diff
@@ -0,0 +1,77 @@
...
+        run: git push origin ${{ steps.extract_branch.outputs.branch }} --follow-tags --force
+
+      - name: Build
+        run: yarn build-dist
+
+      - name: License report
+        run: yarn license-report-dist
+
+      - name: Compare artifact package version
+        run: .github/scripts/compare-artifact-package-versions.sh
```
karthikbtf, line 60: We could skip this perhaps. Remnant from an older build approach. If a version already exists, that's fine, the npm publish will fail. Dev will be notified of a failed build and can follow up. Otherwise if for whatever reason the publish was skipped.. dev might be waiting endlessly on that 'new npm version published' email. @charlie-tf thoughts on removing this from all builds?

karthikbtf (commented):
This is a great productivity improvement 🎉

package.json
charlie-tf: Just tested locally and if `package.json` has `"version": "2.3.59-rc.3"` then `npm version patch` will indeed change it to `"version": "2.3.59"`

nickgraz: I confirmed that it will remove the RC bit.

```
⚡ npm version patch --git-tag-version false
v2.3.59
```

.github/workflows/publish-rc.yaml
charlie-tf: I think that your reasoning is sound. I'd be good with removing it

nickgraz: If I remove from all builds, should I also remove [this](https://github.com/terraware/web-components/blob/main/.github/workflows/develop.yml#L61) line?

karthikbtf: Yes 👍

nickgraz: Hmm I think we still might want to keep it. Otherwise it will always appear like the build failed. See the latest "Deployment" run for this PR.

karthikbtf (approved):
Awesome!

.github/workflows/publish-rc.yaml
karthikbtf: Oops..

karthikbtf: You are right, sorry about that. Yes, best to revert the last commit. Keep things as they were for all builds.

karthikbtf (approved):
Thanks, looks good 👍

PR #434, opened 2023-09-13
SW-4199 Intersection Observer story

nickgraz:
- Implement basic story with `react-intersection-observer` to show how it works in action

It is a pretty basic example, and implements the `react-multi-carousel` since it is a heavy component and has visible lag in rendering the photos. Under the hood the DOM is updated pretty much instantaneously so using in our PDD tool on a tall page with many sections will seem much snappier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "moment-timezone": "^0.5.43",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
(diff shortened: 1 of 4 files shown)


constanzauanini (approved):
This looks and works great!

nickgraz:
Taking the patch label off this since there isn't anything net-new in the NPM package

PR #437, opened 2023-09-21
SW-4241 Allow disabled table row checkboxes and tooltip titles for the checkbox

karthikbtf:
- added new optional prop func `isSelectable` that returns boolean if a row is selectable
- this can be used to dictate whether a row is selectable or not, the component does the right math to show selected-all logic
- disabled selections will show a disabled checkbox
- also add support for showing tooltips next to a checkbox (if use-case wants to explain why a row is not selectable)
- added a story to capture this
- this will be used to disable selecting certain users in TW People table

Screen Shot 2023-09-21 at 3 38 31 PM

Table - Is Selectable ⋅ Storybook 2023-09-21 15-38-13

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.3.65",
+  "version": "2.3.66-rc.0",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "test-coverage": "cp tsconfig.json tsconfig.json.bak && react-scripts test . --coverage --coverageDirectory='docs/coverage' && mv tsconfig.json.bak tsconfig.json",
     "eject": "react-scripts eject",
     "storybook": "storybook dev",
+    "storybook-legacy": "NODE_OPTIONS=--openssl-legacy-provider storybook dev",
     "build-storybook": "storybook build -s public",
     "build-dictionary": "./style-dictionary/generate.sh",
     "transpile": "rm -rf dist && NODE_ENV=production babel src/ --extensions '.tsx,.ts' --out-dir dist --copy-files --no-copy-ignored && cp package.json dist/",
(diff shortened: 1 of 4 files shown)


karthikbtf:
@chudo76 @hannah-cai lmk your thoughts on this approach (for when certain rows in a table are not selectable)
Example: we don't want to select the 'Terraware Contact' user in Terraware org users table

karthikbtf:
publish rc

constanzauanini (approved):
Code looks correct and working correctly in Vercel story

karthikbtf:
> Code looks correct and working correctly in Vercel story

Thanks @constanzauanini I’ll also wait for design team to comment in case they prefer something else.

chudo76:
> @chudo76 @hannah-cai lmk your thoughts on this approach (for when certain rows in a table are not selectable) Example: we don't want to select the 'Terraware Contact' user in Terraware org users table

Sorry, just seeing this after commenting on [SW-4239](https://terraformation.atlassian.net/browse/SW-4239). The information icon looks a bit off. The way I was thinking about this is that you would have bulk action buttons for any selected row be the thing that's disabled instead of the row's checkbox. This gives us a bit more room to work with (ex. we can prevent users from "Remove" but still allow "Edit") and provide more granular tooltips:
image

chudo76:
(I've done something similar in prior apps):

image

karthikbtf:
> > @chudo76 @hannah-cai lmk your thoughts on this approach (for when certain rows in a table are not selectable) Example: we don't want to select the 'Terraware Contact' user in Terraware org users table
> 
> Sorry, just seeing this after commenting on [SW-4239](https://terraformation.atlassian.net/browse/SW-4239). The information icon looks a bit off. The way I was thinking about this is that you would have bulk action buttons for any selected row be the thing that's disabled instead of the row's checkbox. This gives us a bit more room to work with (ex. we can prevent users from "Remove" but still allow "Edit") and provide more granular tooltips: image

Ok sounds good, essentially disable Remove if TF Contact is selected, regardless of whether there are more selections with removable users.

PR #463, opened 2024-01-20
SW-4565 / SW-3514 Make table headers title case and don't overflow

alexTerraformation:
Screenshot 2024-01-19 at 4 00 59 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.5.2",
+  "version": "2.5.3-rc.0",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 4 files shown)


alexTerraformation:
publish rc

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.5.2",
+  "version": "2.5.3-rc.0",
```
karthikbtf, line 3: does this need to be undone or ok as is?

karthikbtf (approved):
🎉

package.json
alexTerraformation: Yes, oops.  Thank you!

alexTerraformation: @nickgraz I've not used this before.  Will this get replaced with the new version upon release in NPM?

nickgraz: OK as is, the next version will be `2.5.3`, which is tagged when this is merged in. `2.5.3-rc.0` is the first release candidate for `2.5.3` and the version here will get overwritten when the branch is merged.

nickgraz (approved):

PR #469, opened 2024-02-08
Fix MultiSelect-within-Popover functionality

nickgraz:
The previous PR that added this functionality did not work as expected, and required the user to double click out of the Popover if they had focused and blurred the MultiSelect _within_ the Popover.

The implementer of this MultiSelect-within-Popover-with-desired-blur/closing-behavior unfortunately needs to implement a cheeky setTimeout for the functionality to work as expected. This is because the state of the component updates from the onBlur event resolution before the onClick event resolves, resulting in an impossible (at least through my many attempts) situation to achieve the desired effect.

### DIFF
--- a/src/stories/MultiSelect.stories.tsx
+++ b/src/stories/MultiSelect.stories.tsx
@@ -56,16 +56,7 @@ const WithParentOptionsVisibilityControlTemplate: Story<MultiSelectProps<number,
   const [selectedOptions, setSelectedOptions] = useState<number[]>([]);
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const [optionsVisible, setOptionsVisible] = useState(false);
-  /*
-    Admittedly, this is a bit confusing. Because the onBlur event within the MultiSelect bubbles up _before_ the
-    onClose event in the Popover, we need to store, within this component's state, whether or not the options _were_
-    visible before the blur event occurred. I tried many, many configurations where setOptionsVisible was passed into
-    the MultiSelect, but the onBlur would still fire before the child component could tell the parent component that
-    the options were no longer visible (using setOptionsVisible), which resulted in the options being close, the parent
-    component re-rendering, and then the onClose handler would think that the options aren't visible, so the Popover
-    should be closed, which resulted in not achieving the goal.
-   */
-  const [optionsWereVisibleBeforeBlur, setOptionsWereVisibleBeforeblur] = useState(false);
+  const [shouldClose, setShouldClose] = useState(false);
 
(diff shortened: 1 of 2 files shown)


nickgraz:
Current dependencies on/for this PR:
* `main`
  * **PR #469** <a href="https://app.graphite.dev/github/pr/terraware/web-components/469?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/web-components/469?utm_source=stack-comment).

src/stories/MultiSelect.stories.tsx
```diff
@@ -56,16 +56,7 @@ const WithParentOptionsVisibilityControlTemplate: Story<MultiSelectProps<number,
   const [selectedOptions, setSelectedOptions] = useState<number[]>([]);
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const [optionsVisible, setOptionsVisible] = useState(false);
-  /*
-    Admittedly, this is a bit confusing. Because the onBlur event within the MultiSelect bubbles up _before_ the
```
nickgraz, line 60: @karthikbtf let me know what you think. This is less than ideal but I was unable to figure out another way to achieve the functionality given the way the MUI Popover works. Thankfully the `setTimeout` is something the implemented in the parent, and not within the `MultiSelect`. 

If this isn't acceptable, I'm going to have to say we hold off on this functionality for now because I haven't been able to figure out a better solution and I've already spent a bit too long on this.

karthikbtf: @nickgraz thanks for revisiting, didn't intend for you to spend more time on this with my previous comment.
Interesting interactions/event problem.
This current approach is fine, the previous approach seemed cleaner to me. 
Current approach seems to keep options open even after selecting an item in the dropdown list.
I agree let's not spend much time on making it work.

nickgraz: Ah, good catch. I had to ignore the click capture unless it was on the Popover backdrop. I agree the other solution was cleaner, but unfortunately it had a bug that was a symptom of the two events firing sequentially.

karthikbtf: Got it, this sounds like a good solution to stick with.

karthikbtf: Thanks for fixing the click issue 🙌

karthikbtf (approved):
🎉 Thanks for figuring this out 🎉

PR #480, opened 2024-03-12
Apply new prettier config to entire repo

nickgraz:


karthikbtf (approved):
🎉

src/style-dictionary-dist/TerrawareTheme.ts
```diff
@@ -371,187 +371,193 @@ type ThemeDictionary = {
 };
 
 const TerrawareTheme: ThemeDictionary = {
-  "palette": {
```
joshbuchea, line 374: Probably want to ignore everything in `src/style-dictionary-dist/` since it's generated.

karthikbtf: Good catch. 
My 2 cents, since this is generated by us and within our control, maybe it's ok to format. Or have the theme generator run the prettier on those files (that could be overkill).

nickgraz: Good catch, I put it back.

PR #481, opened 2024-03-13
SW-5031: In Participant view, 1c. Parent/Subsidiary information, no upload option after CANCEL

joshbuchea:
This PR includes changes to fix an issue that was preventing file uploads for deliverables in some scenarios.

### Before

https://github.com/terraware/web-components/assets/1474361/e31256af-79df-4999-9a14-075e694f1aa6

### After

https://github.com/terraware/web-components/assets/1474361/1e52eaa4-4ffd-4cf7-a172-ec2c92aff692

### DIFF
--- a/src/components/FileChooser/index.tsx
+++ b/src/components/FileChooser/index.tsx
@@ -119,6 +119,10 @@ export default function FileChooser(props: FileChooserProps): JSX.Element {
     if (event.currentTarget.files) {
       addFiles(event.currentTarget.files);
     }
+    if (event.currentTarget.value) {
+      // reset the input value to allow the same file(s) to be selected again
+      event.currentTarget.value = '';
+    }
   };
 
   return (
--- a/src/components/FileChooser/index.tsx
+++ b/src/components/FileChooser/index.tsx
@@ -119,9 +119,9 @@ export default function FileChooser(props: FileChooserProps): JSX.Element {
     if (event.currentTarget.files) {
       addFiles(event.currentTarget.files);
     }
-    if (event.currentTarget.value) {
+    if (inputRef?.current?.value) {
       // reset the input value to allow the same file(s) to be selected again
-      event.currentTarget.value = '';
+      inputRef.current.value = '';
     }
   };
 


karthikbtf (commented):
Thanks for taking a look at this sad bug. A few comments.

src/components/FileChooser/index.tsx
```diff
@@ -119,6 +119,10 @@ export default function FileChooser(props: FileChooserProps): JSX.Element {
     if (event.currentTarget.files) {
       addFiles(event.currentTarget.files);
     }
+    if (event.currentTarget.value) {
+      // reset the input value to allow the same file(s) to be selected again
+      event.currentTarget.value = '';
```
karthikbtf, line 124: Can we try another approach like setting the value for the hidden input (see line 151).
Ideally we should not be editing the event object passed into this function. 
@nickgraz may have some options up his sleeve.

karthikbtf (commented):
Thanks for taking a look at this sad bug. A few comments.

src/components/FileChooser/index.tsx
joshbuchea: @karthikbtf sure, we should be able to use the `inputRef` instead.

joshbuchea: There are definitely several approaches that would work here, will try another approach.

joshbuchea: Line `151` seems to reference `onChange={fileChosen}`. The value was and is still being updated within the onChange handler, `fileChosen`. I updated the approach to set the value using the `inputRef` rather than `event.currentTarget`.

Alternatively, we could also force the `input` element to re-render by setting a key with something like `key={fileChooserId}` and `const [fileChooserId, IncrementFileChooserId] = useReducer(x => x + 1, 0);` or use another trick to force react to rerender a component/element.

joshbuchea: We can also force a rerender of the `FileChooser` from the `terraware-web` side, but I thought it felt cleaner to reset the file input here. But happy to adjust the approach, just let me know what you all would prefer.

nickgraz: I think rerendering is the cleanest approach, consider the `FileChooser` a "clean" input that is discarded (and rerendered) if changes are discarded or there is an error. You can wrap the component in a `useMemo` in `terraware-web` and add a dependency that you know will change when an upload fails or is cancelled. 

Also, modifying a ref element is not recommended [in the docs](https://react.dev/reference/react/useRef#caveats), so even if rerendering isn't the right approach we should try something else.

joshbuchea: Sounds like a plan, will make the changes over on the `terraware-web` side. Good callout re: not modifying a ref element.

joshbuchea: The only thing with making the change over in `terraware-web` (I'll still make the change there), this component will always require rerendering from the parent when the same file(s) are selected subsequently. It seems like that issue should be fixed within this component since it's a quirk of working with the html `input` element.

karthikbtf: @joshbuchea might setting `value=''` prop on the hidden input help solve this problem? We will still get the onChange callback, which is all we want. The value will always be cleared allowing for a repeat file selection. Just curious if you can give that a shot.

nickgraz: @karthikbtf that was it! One liner to fix https://github.com/terraware/web-components/pull/482

joshbuchea: I thought it felt like it should be fixed in this repo, great work!

PR #488, opened 2024-03-20
SW-5091 Truncated multi line text

nickgraz:
- Add `TruncatedTextArea` component which renders conditionally within the `Textfield` if we are using `type='textarea'` and a truncation config is passed in

- Since we can't predict how the text will wrap on the page, we need to measure it on the page, as a result, the truncation config needs to have a `maxHeight` which is cleanly divisible by the line height for the value text, otherwise you will see things get cut off.
- I decided this was more simple than measuring and then figuring out what the number of lines is, and then slicing the value. Hopefully it is OK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.10.6-rc.0",
+  "version": "2.10.6-rc.1",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.10.5",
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#488** <a href="https://app.graphite.dev/github/pr/terraware/web-components/488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @nickgraz and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

nickgraz:
publish rc

nickgraz:
publish rc

PR #499, opened 2024-04-04
SW-4662 Avoid flash of dropdown menu before it's position is calculated/repositioned

karthikbtf:
- use state + css to avoid showing the dropdown until it is positioned


karthikbtf:
publish rc

nickgraz (approved):
Nice

karthikbtf:
publish rc

PR #500, opened 2024-04-05
Align table header label and icons to the right when column is right aligned

constanzauanini:
Previously the label was aligned to right, but not the drag, sort and tooltip icons. Reorder all icons correctly when column is right-aligned

Sort icon
Screenshot 2024-04-05 at 10 45 14

Drag icon
Screenshot 2024-04-05 at 10 57 39


nickgraz (approved):
Nice work

PR #507, opened 2024-04-09
SW-4662 Don't resize dropdown during scroll

karthikbtf:
- the resize happens as part of existing code when it repositions the menu during a window resize
- skip the resize if this is part of scroll


constanzauanini (approved):
Thanks for the fix!

karthikbtf:
> Thanks for the fix!

Thanks for helping with the validation!!

PR #514, opened 2024-04-17
SW-5218 Fix datepicker adapter for luxon

karthikbtf:
- the x-date-pickers luxon adapter is off mui5.x which has bugs
- specifically uses `instanceof DateTime` to check for a luxon dates, instanceof does not work across namespaces (npm package scope vs root terraware-web scope)
- this was preventing min/max date props from working across namespaces
- same for setting the value of DatePicker without a toISO conversion
- moved to @date-io/luxon which uses luxon's isDateTime function to compare dates
- this fixes all the issues previously observed (cannot delete characters, min/max dates not working)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "2.16.2",
+  "version": "2.16.2-rc.0",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


karthikbtf:
publish rc

tommylau523 (approved):

karthikbtf:
publish rc

PR #548, opened 2024-05-15
SW-5312: Update styles for Tab Group/Tab component (follow-ups)

joshbuchea:
This PR includes some follow-up style updates that weren't included in the previous PR.

### DIFF
--- a/src/components/Tabs/index.tsx
+++ b/src/components/Tabs/index.tsx
@@ -44,6 +44,7 @@ const Tabs = ({ tabs, onTabChange, activeTab }: TabsProps): JSX.Element => {
       alignItems: 'center',
     },
     '& .MuiTab-iconWrapper': {
+      fill: '#7F775B',
       marginBottom: 0,
       marginRight: theme.spacing(1),
     },
--- a/src/components/Tabs/index.tsx
+++ b/src/components/Tabs/index.tsx
@@ -28,6 +28,7 @@ const Tabs = ({ tabs, onTabChange, activeTab }: TabsProps): JSX.Element => {
     color: '#7F775B',
     fontSize: '16px',
     fontWeight: 600,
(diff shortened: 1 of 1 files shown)


src/components/Tabs/index.tsx
```diff
@@ -43,6 +44,7 @@ const Tabs = ({ tabs, onTabChange, activeTab }: TabsProps): JSX.Element => {
       alignItems: 'center',
     },
     '& .MuiTab-iconWrapper': {
+      fill: '#7F775B',
```
chudo76, line 47: I think we want to avoid hard-coded hex values. I believe we want `theme.palette.TwClrIcnSecondary as string` here.

src/components/Tabs/index.tsx
chudo76, line 1: Could we change [line 52](https://github.com/terraware/web-components/blob/7b39325511dc2375b36a74ce485955ebab3df447/src/components/Tabs/index.tsx#L52):
`fill: theme.palette.TwClrTxtBrand as string` to
`fill: theme.palette.TwClrIcnBrand as string`

🙏

src/components/Tabs/index.tsx
joshbuchea: I'll update this to use the color token, but heads up that the hex value is also incorrect for this color token.

chudo76: I'm seeing `TwClrIcnSecondary`/`tw-clr-icn-secondary` mapping to `#7F775B` on Storybook. I think that's correct?

chudo76 (approved):
TY!

src/components/Tabs/index.tsx
joshbuchea: You're correct, I'm not sure what I thought I saw 😅

PR #558, opened 2024-05-22
SW-5387 Fix tables are missing right padding on all pages

constanzauanini:
Remove use of useStyles


constanzauanini:
publish rc

tommylau523 (approved):

PR #559, opened 2024-05-22
SW-5387 Fix tables are missing right padding on all pages

constanzauanini:
Remove use of useStyles


tommylau523 (approved):
Looks good to me!

PR #565, opened 2024-05-28
SW-5420 Add style to table columns

constanzauanini:
Add style prop so we can set width to the columns of the table


nickgraz (approved):
Nice work! Glad to see we're moving away from `useStyles`

PR #579, opened 2024-06-05
SW-5494 Update theme builder to work with ESLint

alexTerraformation:
ESList complains about the theme that gets built.  Update the builder script so that it's happy.


constanzauanini (approved):
Tested locally and it's working. Thanks for fixing this!

PR #582, opened 2024-06-05
SW-5443 | SW-5442 | SW-5444 Update table styles

alexTerraformation:
Add striped row background. Increase cell font-size to 16px. Add hover and selected row styles.


alexTerraformation:
publish rc

src/style-dictionary-dist/TerrawareTheme.ts
```diff
@@ -369,193 +371,188 @@ type ThemeDictionary = {
...
+    'TwClrBrdrWarningActive': '#815424',
+    'TwClrBrdrWarningHover': 'linear-gradient(rgb(129 84 36 / 50%), rgb(129 84 36 / 50%)), linear-gradient(#A36A2D, #A36A2D)',
+    'TwClrIcnOnWarning': '#FFFFFF',
+    'TwClrIcnWarning': '#A36A2D',
+    'TwClrTxtOnWarning': '#FFFFFF',
+    'TwClrTxtWarning': '#A36A2D',
+    'TwClrBgAccent': '#EF7047',
+    'TwClrTxtAccent': '#EF7047',
+    'TwClrTxtOnAccent': '#FFFFFF'
+  }
```
nickgraz, line 555: Do you know which colors were removed?

nickgraz (approved):
Looks good to me, hopefully the missing colors are either not used or throws compilation errors in `terraware-web`

src/style-dictionary-dist/TerrawareTheme.ts
alexTerraformation: I don't think any colors were removed.  If you're just looking at line numbers then there were some unused lines at the top of the file that were removed because they were causing ESLint to trip

alexTerraformation: I don't know.  Since it's autogenerated from the style dictionary I suspect it was design team removing unused colors.  I can double-check with Chudo.

PR #601, opened 2024-06-18
Truncate table cell

constanzauanini:
The root in the classes object for typography meant that the style should be applied to the MUI root class. With MUI 5 we don't need to do that, just set the max-width to the sx object


joshbuchea (approved):
Ok good to know that we shouldn't use `root` in `sx` props, thanks for fixing this 🙌

PR #606, opened 2024-06-18
Fix table header padding

constanzauanini:
When adding the "Density settings" the padding of the headers were disabled, but this was not necessary. Revert it to have the padding back and remove the styles added after for fixing that issue


joshbuchea (approved):
Nice! 🙌 🚢

PR #607, opened 2024-06-19
SW-5538: Table column sorting for Species table for Species List Deliverable doesn't seem to work

joshbuchea:
This PR includes a change to support sorting by nested row properties in the `Table` component.

### DIFF
--- a/src/components/table/sort.ts
+++ b/src/components/table/sort.ts
@@ -42,9 +42,9 @@ function isEmptyValue(value?: string): boolean {
   }
 }
 
-function descendingNumComparator<T>(aValue: T, bValue: T): number | null {
-  const aNumValue = Number(aValue);
-  const bNumValue = Number(bValue);
+function descendingNumComparator<T>(a: T, b: T): number | null {
+  const aNumValue = Number(a);
+  const bNumValue = Number(b);
 
   if (!isNaN(aNumValue) && !isNaN(bNumValue)) {
     return bNumValue - aNumValue;
--- a/src/components/table/sort.ts
+++ b/src/components/table/sort.ts
@@ -1,13 +1,15 @@
-export function descendingComparator<T>(a: T, b: T, orderBy: keyof T, order: SortOrder): number {
-  // first attempt to parse into a numeric value and compare
-  const numCompare = descendingNumComparator(a, b, orderBy);
-  if (numCompare !== null) {
-    return numCompare;
-  }
+export function descendingComparator<T>(a: T, b: T, orderBy: keyof T | string, order: SortOrder): number {
+  const getValue = (obj: any, path: string) => {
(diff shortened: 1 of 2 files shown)


joshbuchea:
I believe these changes retain the previous functionality while adding support for nested properties, but please let me know if you think I missed anything 🙏

tommylau523 (approved):

src/components/table/sort.ts
```diff
@@ -40,9 +42,9 @@ function isEmptyValue(value?: string): boolean {
   }
 }
 
-function descendingNumComparator<T>(a: T, b: T, orderBy: keyof T): number | null {
-  const aNumValue = Number((a[orderBy] ?? '0.0') as string);
-  const bNumValue = Number((b[orderBy] ?? '0.0') as string);
+function descendingNumComparator(aValue: any, bValue: any): number | null {
```
nickgraz, line 45:
```suggestion
function descendingNumComparator(aValue: unknown, bValue: uknown): number | null {
```

`unknown` is better to use than `any` if you are unsure what something is. `any` allows anything, whereas `unknown` forces type checking where necessary.

src/components/table/sort.ts
```diff
@@ -1,13 +1,15 @@
-export function descendingComparator<T>(a: T, b: T, orderBy: keyof T, order: SortOrder): number {
-  // first attempt to parse into a numeric value and compare
-  const numCompare = descendingNumComparator(a, b, orderBy);
-  if (numCompare !== null) {
-    return numCompare;
-  }
+export function descendingComparator<T>(a: T, b: T, orderBy: keyof T | string, order: SortOrder): number {
+  const getValue = (obj: any, path: string) => {
+    return path.split('.').reduce((acc, part) => acc && acc[part], obj);
+  };
```
nickgraz, line 4: Not really necessary, but here's how you could rewrite it if you want to be type safe and avoid using `any`. 

```suggestion
type UnknownObject = { [key: string]: unknown | UnknownObject };

export function descendingComparator<T extends UnknownObject>(
  a: T,
  b: T,
  orderBy: keyof T | string,
  order: SortOrder
): number {
  const getValue = (obj: unknown, path: string): unknown =>
    path
      .split('.')
      .reduce(
        (acc: unknown | UnknownObject, part: string): unknown | UnknownObject => acc && (acc as UnknownObject)[part],
        obj
      );
```

nickgraz (approved):
This is fine, a couple suggestions for improving the typescript usage

nickgraz (approved):

joshbuchea:
Oops! I meant to merge this yesterday 😅 Thank you for the reviews and suggestions 🙌 💯

joshbuchea:
I realized the behavior to split column key strings by dot (".") should be optional so it doesn't break other tables using dot notation in column keys, so I added an optional `sortSplitDots` param to the `Table` component.

PR #625, opened 2024-07-31
SW-5701 Add title property to all table cells

constanzauanini:


nickgraz (approved):
Looks good. Can you change the label from `patch` to `minor`?

PR #630, opened 2024-08-21
SW-5825: Can't type accents in Questionnaire Deliverable inputs

joshbuchea:
This PR includes a change to add a `defaultValue` prop to the `Textfield` component.

### DIFF
--- a/src/components/Textfield/Textfield.tsx
+++ b/src/components/Textfield/Textfield.tsx
@@ -23,6 +23,7 @@ export interface TruncateConfig {
 export interface Props {
   autoFocus?: boolean;
   className?: string;
+  defaultValue?: string | number;
   disabled?: boolean;
   disabledCharacters?: string[];
   display?: boolean;
@@ -58,6 +59,7 @@ export default function TextField(props: Props): JSX.Element {
     autoFocus,
     label,
     className,
+    defaultValue,
     disabled,
     disabledCharacters,
     display,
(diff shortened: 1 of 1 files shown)


src/components/Textfield/Textfield.tsx
```diff
@@ -163,7 +175,8 @@ export default function TextField(props: Props): JSX.Element {
             {iconLeft && <Icon name={iconLeft} className='textfield-value--icon-left' />}
             <input
               autoFocus={autoFocus}
-              value={type === 'number' ? value : value || ''}
+              defaultValue={defaultValue}
+              value={type === 'number' || defaultValue !== undefined ? value : value || ''}
```
constanzauanini, line 179: I don't get why `defaultValue !== undefined` condition is added here

joshbuchea: When `defaultValue` is set, `value` should be `undefined`. Previously, `value` defaulted to an empty string (`''`) when not set. The new `defaultValue !== undefined` check ensures `value` is passed correctly when `defaultValue` is used.

constanzauanini (approved):
Changes to add default values looks correct. I don't see the relationship with the accents though

joshbuchea:
Closing in favor of https://github.com/terraware/terraware-web/pull/3064

joshbuchea:
@constanzauanini thanks for the review. I tried an approach to resolve SW-5825 that involved updating `DeliverableVariableDetailsInput` to be an uncontrolled component, which appeared to resolve the cursor position reset issues, but I later discovered generated a new React warning.

I ended up going with a different approach that didn't require this change.

PR #632, opened 2024-08-24
SW-5781: Add default file name (project + template) when people create a new document

joshbuchea:
This PR adds an `onFocus` prop to the `Textfield` component to trigger some code when the "Document Name" field is focused in the Add New Document view.


constanzauanini:
Is the on focus implementation already added?

constanzauanini (approved):

PR #633, opened 2024-08-27
Add renderOption prop to autocomplete

constanzauanini:
Autocomplete renders a < ul > with < li > children. By default those < li > children have the label of the option as key, and sometimes labels could be the same, so it generates this error:

`Encountered two children with the same key, ``. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

MUI Autocomplete has a renderOption prop to change that default behaviour, so we need to add it to our Autocomplete component to allow using it


nickgraz (approved):

joshbuchea:
💯

PR #640, opened 2024-10-14
Add date time picker option

constanzauanini:


tommylau523 (approved):

tommylau523 (commented):
No opinion either way but would it make kore sense to have a separate component for DateTimePicker, instead of using a prop to decide?

Another feature we may need for the time picker is to pick a timezone, since our DB mostly use the datetime with zone offset format. Could be handy for scheduling events for accelerator for example.

constanzauanini:
> No opinion either way but would it make kore sense to have a separate component for DateTimePicker, instead of using a prop to decide?
> 
> Another feature we may need for the time picker is to pick a timezone, since our DB mostly use the datetime with zone offset format. Could be handy for scheduling events for accelerator for example.

I started by doing two separate component, but they were exactly the same, that's why I choose to add a prop. Also the datepicker already have a defaultTimeZone prop

PR #693, opened 2025-09-03
Used new markdown library

tommylau523:
Moved to use a pure js library instead of a React library to simplify some packaging and transpiling issues. 

This also created a new Markdown component to be used separately. Although this is not exported directly yet.


joshbuchea (approved):
Changes look good!

PR #765, opened 2026-05-27
Fix generate-assets script

constanzauanini:
- PR #735 bumped @trivago/prettier-plugin-sort-imports from v4 (CJS) to v6 (ESM with top-level await).
- Prettier 2 (used by @svgr/plugin-prettier@8) loads plugins via require(), which on Node 20.19+ throws ERR_REQUIRE_ASYNC_MODULE when asked to require an ESM module with top-level await — so svgr crashes before producing any output

.svgrrc.cjs configures svgr to:
  - runtimeConfig: false — stops svgr's prettier from auto-discovering the repo's .prettierrc. That file declares the trivago plugin in its "plugins" field, so any prettier instance that reads it will then try to require() the plugin (which is the call that crashes inside svgr's bundled prettier 2)
  - prettierConfig: {...} — supply the same formatting options inline so generated files keep the project's single-quote
   style.

### DIFF
--- a/.svgrrc.cjs
+++ b/.svgrrc.cjs
@@ -1,0 +1,11 @@
+module.exports = {
+  runtimeConfig: false,
+  prettierConfig: {
+    trailingComma: 'es5',
+    semi: true,
+    singleQuote: true,
+    jsxSingleQuote: true,
+    tabWidth: 2,
+    printWidth: 120,
+  },
+};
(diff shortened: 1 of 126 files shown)


ice1080 (approved):
It's a bummer that these svg components won't follow the rest of the prettier config for the newline after imports, but obviously not a big deal

PR #771, opened 2026-07-22
SW-6258 Focus text field when opening options so keyboard navigation always works

constanzauanini:


ice1080 (approved):
It may be worth using `yalc` or similar to help with development of changes in web-components, if this ends up not fixing what you're hoping!

constanzauanini:
> It may be worth using `yalc` or similar to help with development of changes in web-components, if this ends up not fixing what you're hoping!

Yes, it did fix it, that's why I merged. Then I realized that in some cases it was not working because of the focus

PR #789, opened 2026-07-30
Allow PhotoChooser to have an existing list of images

ice1080:
Update to allow PhotoChooser to have a list of existing images as well as the list of new images. 
Update AnnotationEditPane to pass existing images to this.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

---


ice1080:
publish rc

constanzauanini (approved):
I think in web-components we are showing existing images in some places, so maybe that code could be updated to use it like this

ice1080:
> I think in web-components we are showing existing images in some places, so maybe that code could be updated to use it like this

I'll do that in a follow up PR!

PR #798, opened 2026-08-03
Fix the ground height for VR

ice1080:
The ground height of the model was set close to where the viewer's headset was when inside VR. Set this to LocalFloor so that it feels more natural.

---


ice1080:
publish rc

tommylau523 (approved):

PR #803, opened 2026-08-03
SW-8724 Fix splat culling in VR

ice1080:
Playcanvas culls splats outside of the camera's FOV so that it only renders what's visible. 
When in VR, this was culling behind the fixed camera from the desktop instead of behind the headset's camera.
Use the unified splat prop to cull behind the headset instead.

Also upgrade playcanvas as part of this to make sure we're using the latest version of their unified script.

---


src/components/VirtualWalkthrough/VirtualWalkthroughViewer.tsx
```diff
@@ -270,29 +272,33 @@ const VirtualWalkthroughViewer = ({
...
         <Script script={XrControllers} enabled={!isEdit} />
-        <Script script={TfXrNavigation} enabled={!isEdit} enableTeleport={false} />
-        <Script
-          script={AutoRotator}
-          enabled={!isEdit && autoRotate && !viewingAnnotation}
-          startDelay={0.5}
-          restartDelay={3}
-          startFadeInTime={0.5}
-        />
+        <Script script={TfXrNavigation} enabled={!isEdit} enableTeleport={true} />
```
constanzauanini: The enableTeleport change is needed too?

constanzauanini (approved):

ice1080:
publish rc

src/components/VirtualWalkthrough/VirtualWalkthroughViewer.tsx
ice1080: Actually I need to test a few things. There's a chance that our whole wrapper script is no longer needed at all. I can't seem to find why this whole script was necessary in the first place

ice1080:
publish rc

ice1080:
publish rc

src/components/VirtualWalkthrough/VirtualWalkthroughViewer.tsx
ice1080: Good callout. I had forgotten that this was only for AR (e.g. in the phone). Added a comment instead. There's still a chance that we don't need our own script for it if they fixed the teleport bug, but I'll try that in a separate PR.

PR #816, opened 2026-08-07
SW-8744 Avoid closing modal on mobile in specific cases

constanzauanini:


ice1080 (approved):
I've noticed this and thought it should be fixed but never dived into the code. Thanks for fixing!

PR #817, opened 2026-08-07
Fix PhotosCarousel default import of react-multi-carousel

ice1080:
Since we began publishing dist/ as native ES modules (eb23d57f5),
`import Carousel from 'react-multi-carousel'` resolves to the package's whole
`module.exports` object rather than the component, so consumers builds fail 
at runtime when PhotosCarousel renders.

react-multi-carousel is CommonJS compiled from TypeScript, so it ships the
`exports.default` + `__esModule` shape that relies on a compiler-emitted
interop helper to unwrap. Our dist Babel config applies no module transform,
so no such helper is emitted, and `"type": "module"` puts both bundlers into
Node-spec ESM mode where the default import of a CommonJS module is defined
to be `module.exports` verbatim — `__esModule` is deliberately ignored.

Unwrap the nested default at the import site. The `??` keeps this correct
under bundlers that do still apply the legacy interop.

`Carousel` was also used as a type for the ref, which a const binding cannot
provide, so derive `CarouselInstance` via `InstanceType` for `useRef`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terraware/web-components",
-  "version": "4.2.25",
+  "version": "4.2.26-rc.0",
   "author": "Terraformation Inc.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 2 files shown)


ice1080:
This bug was complex enough that I decided to leave the AI generated description since it explains things pretty well.

ice1080:
publish rc

ice1080:
publish rc

tommylau523 (approved):

PR #818, opened 2026-08-07
Build the published package with Rollup

ice1080:
Replaces the per-file `babel --out-dir` transpile with a Rollup build that still emits one ES module per source file (preserveModules), so the published layout and deep-import paths are unchanged.

Removes publishing of test files since those aren't needed.

This was tested in storybook locally against every component, and also against both terraware-web and virtual-walkthrough-story to ensure general functionality.


ice1080:
publish rc

ice1080:
publish rc

ice1080:
publish rc

tommylau523 (approved):

PR #820, opened 2026-08-07
Scale walkthrough camera targets by the scene scaleFactor

ice1080:
The camera rig sits outside content-root, which scales the scene by
scaleFactor, but the focus and eye points handed to WalkthroughCamera.reset
were passed through in scene space. The bounds circle, ground plane and
average camera height were already scaled, so the reset landed the camera a
scaleFactor closer to the middle of the splat than intended - at
scaleFactor 15 that is inside the model.

Convert origin, cameraPosition and annotation coordinates to world space at
the call sites, and report CameraInfo's readout back in scene space so the
poses it is used to author round-trip through the same coordinates.

---


ice1080:
Superseded by #823: coordinates handed to the viewer are world-space now, so there is nothing left to scale at the camera.

